PR #17, opened 2020-04-14
airflow maintenance dag added

tayowonibi:
partly addresses the issue [#5568](https://github.com/elifesciences/issues/issues/5568)

### DIFF
--- a/dags/dag_maintenance.py
+++ b/dags/dag_maintenance.py
@@ -8,7 +8,7 @@ airflow trigger_dag --conf '{"maxDataAgeInDays":30}' airflow-log-cleanup
 """
 import os
 import logging
-from datetime import datetime, timedelta
+from datetime import timedelta
 import dateutil.parser
 import airflow
 from airflow.models import (
@@ -82,9 +82,9 @@ DATABASE_OBJECTS = [
     },
 ]
 
-session = settings.Session()
+SESSION = settings.Session()
 
-default_args = {
+DEFAULT_ARGS = {
     'depends_on_past': False,
     'email_on_failure': True,
     'email_on_retry': False,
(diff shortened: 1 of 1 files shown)


dags/dag_maintenance.py
```diff
@@ -121,18 +121,18 @@ def get_max_data_cleanup_configuration_function(**context):
     context["ti"].xcom_push(key="max_date", value=max_date.isoformat())
 
 
-get_configuration = PythonOperator(
+GET_CONFIGURATION = PythonOperator(
```
de-code: suggestion: what I ended up doing was to move the code to a `create_dag` function so that these variables become local variables. Because it doesn't seem that these are intended to be "constants".

dags/dag_maintenance.py
```diff
@@ -0,0 +1,188 @@
...
+
+    cleanup_op = PythonOperator(
+        task_id='cleanup_' + str(db_object["airflow_db_model"].__name__),
+        python_callable=cleanup_function,
+        params=db_object,
+        provide_context=True,
+        dag=MAINTENANCE_DAG
+    )
+
+    GET_CONFIGURATION.set_downstream(cleanup_op)
```
de-code: BTW from https://airflow.apache.org/docs/stable/concepts.html

> We recommend you setting operator relationships with bitshift operators rather than `set_upstream()` and `set_downstream()`.
> 
> Traditionally, operator relationships are set with the `set_upstream()` and `set_downstream()` methods. In Airflow 1.8, this can be done with the Python bitshift operators `>>` and `<<`.

I find `set_downstream` less intuitive, because if it is a regular `set` operation, then we would replace the previously set operator. `set_upstream` would avoid that problem. But maybe using a list is even better.

de-code (approved):

PR #46, opened 2020-06-03
Google analytics data pipeline

tayowonibi:
Added google analytics data pipeline
Refactors existing code to avoid duplication of functionality


data_pipeline/google_analytics/etl_state.py
```diff
@@ -0,0 +1,41 @@
...
+        data_config: GoogleAnalyticsConfig,
+        default_latest_state_date: str = None,
+) -> str:
+    try:
+        stored_state = download_s3_object_as_string(
+            data_config.state_s3_bucket_name,
+            data_config.state_s3_object_name
+        )
+    except ClientError as ex:
+        if ex.response['Error']['Code'] == 'NoSuchKey':
```
de-code, line 34: comment: the google api has separate "Not Found" exceptions. Maybe this could benefit from one as well. This response checking seems quite low level.
(This is more a comment for potential future improvements)

data_pipeline/utils/__init__.py
```diff
@@ -0,0 +1,57 @@
+
+
+def update_deployment_env_placeholder(
```
de-code: Would probably suggest to move them to a dedicated module. Just "utils" feels slightly to generic. It does seem to be related to configuration.

tests/unit_test/test_ga_etl_state.py
```diff
@@ -0,0 +1,72 @@
...
+
+GA_CONFIG = {
+    'gcpProjectName': 'elife-data-pipeline',
+    'dataPipelineId': 'ga_pipeline_id',
+    'dataset': 'dataset'
+}
+
+
+@pytest.fixture(name="mock_download_s3_json_object_exception")
+def _download_s3_json_object_with_exception():
```
de-code: I would make this fixture get "mock_download_s3_json_object" as an input and set the side_effect. But it may be better to just set the side_effect as part of the test itself. You could still have a method that returns the ClientError. e.g.

```python
mock_download_s3_json_object.side_effect = get_no_such_key_client_error()
```

Alternatively that could even be a constant, e.g. NO_SUCH_KEY_CLIENT_ERROR

tests/unit_test/test_ga_etl_state.py
```diff
@@ -0,0 +1,72 @@
...
+    with patch.object(
+            etl_state_module, 'download_s3_object_as_string',
+            side_effect=ClientError(operation_name='GetObject',
+                                    error_response=client_error_response)
+    ) as mock:
+        yield mock
+
+
+@pytest.fixture(name="mock_download_s3_json_object")
+def _download_s3_json_object_():
```
de-code: probably don't need the underscore at the end.

tests/unit_test/test_ga_pipeline.py
```diff
@@ -0,0 +1,165 @@
...
+            {'ga_userType': 'Returning Visitor', 'date_range': '0',
+             'ga_sessions': '12798', 'ga_sessionDuration': '2215033.0'}
+        ]
+
+        assert response == expected_data
+
+
+class TestETLGA:
+    # pylint: disable=too-many-arguments
+    def test_should_etl_ga(
```
de-code, line 135: minor: naming is hard. I guess this just tests that it is not failing and calls the right methods (without checking what is passed).

dags/google_analytics_data_import_pipeline.py
```diff
@@ -0,0 +1,132 @@
...
+    )
+
+
+def google_analytics_etl(**kwargs):
+    externally_triggered_parameters = kwargs['dag_run'].conf or {}
+    external_trigger_conf_dict = externally_triggered_parameters.get(
+        ExternalTriggerConfig.GA_CONFIG
+    )
+    dag_context = kwargs["ti"]
+    multi_keyword_extract_conf_dict = (
```
de-code: name may need to be updated

data_pipeline/google_analytics/ga_pipeline.py
```diff
@@ -0,0 +1,180 @@
...
+            add_provenance(
+                transform_response_to_bq_compatible_record(
+                    response
+                ),
+                ga_config.import_timestamp_field_name,
+                current_timestamp_as_string,
+                ga_config.record_annotations
+            )
+        )
+        with NamedTemporaryFile() as temp_file:
```
de-code: same as https://github.com/elifesciences/data-hub-ejp-xml-pipeline/pull/40#discussion_r433135539

data_pipeline/google_analytics/ga_pipeline.py
```diff
@@ -0,0 +1,180 @@
...
+    ]
+    metrics = [
+        {
+            'expression': metric_name
+        } for metric_name in ga_config.metrics
+    ]
+    page_token = None
+    while True:
+
+        response = analytics.get_report(
```
de-code: I would recommend moving this out to a page generator function. That way this method becomes shorter and it will be easier to see what the stop condition is. And perhaps it's also worth moving the processing of a response to a separate method.

e.g.

```
paged_report_responses = iter_get_report_pages(...)
for paged_report_response in paged_report_responses:
    process_paged_report_response(paged_report_response)
```

data_pipeline/google_analytics/ga_pipeline.py
```diff
@@ -0,0 +1,180 @@
...
+            }
+        }
+
+        yield {
+            **record,
+            **provenance
+        }
+
+
+def write_result_to_file(
```
de-code: the google sheets pipeline seem to have a similar method. Might be worth moving to a common module. I guess the name could reflect that we are writing jsonl.

tests/unit_test/test_ga_pipeline.py
tayowonibi: yeah

data_pipeline/google_analytics/ga_pipeline.py
```diff
@@ -59,6 +59,34 @@ def transform_response_to_bq_compatible_record(
...
+        # api can return multiple reports, however,
+        # this implementation considers single report
+        page_token = (
+            reports[0].get('nextPageToken')
+            if len(reports) == 1 else None
+        )
+        if not page_token:
+            yield response
+            break
+        yield response
```
de-code: minor: I guess the yield could be moved above the if statement (then there would be just one)

Is it possible to have no pages? I guess it would just be a page with no items?

tayowonibi: if there is no next page, `page_token` changes to `None`, and when `page_token` is `None`, then you start all over again from the first page

de-code: I mean, what if there is no content / page at all. Currently we will always yield at least one page.

tayowonibi: you will get a response page with no data content

data_pipeline/google_analytics/ga_pipeline.py
de-code: This comment is marked as resolved. Are we going to replace `NamedTemporaryFile`? I would really love to see it go :-)

tayowonibi: we need some temporary file that we can open for write and/or append, close, and re-open for read, and I can't just figure out how to use ordinary `TemporaryFile` for that, but `NameTemporaryFile`

de-code: Okay, I think I may not explained properly what I am after. I fixed the link to the related conversation. I am suggesting to use `TemporaryDirectory` and then just create a file within that temporary directory.

tayowonibi:
@de-code - I've made all the necessary modifications

de-code (approved):
thank you for doing that

PR #63, opened 2020-06-10
feat(civi-data-export): export civi-crm data

tayowonibi:
part of https://github.com/elifesciences/issues/issues/5633

more tests are being added


sample_data_config/web-api/web-api-data-pipeline.config.yaml
```diff
@@ -20,16 +20,63 @@ webApi:
         - items
   #elifescience profile api
   - dataset: '{ENV}'
-    table: atest_profile_api
+    table: test_people_api
```
de-code, line 23: was this change intentional? now there seem to be two people api configuration items

sample_data_config/web-api/web-api-data-pipeline.config.yaml
```diff
@@ -20,16 +20,63 @@ webApi:
...
+        fieldsToReturn:
+          - external_identifier
+          - tag
+          - group
+          - contact_type
+          - contact_sub_type
+          - last_name
+          - first_name
+          - middle_name
+          - email,source
```
de-code: what does having comma separated values do?

(I am trying to use the config as a starting point to understand the change better)

sample_data_config/web-api/web-api-data-pipeline.config.yaml
```diff
@@ -20,16 +20,63 @@ webApi:
...
         - items
+  # civi export
+  - dataset: '{ENV}'
+    table: civi_data_export
+    stateFile:
+      bucketName: '{ENV}-elife-data-pipeline'
+      objectName: 'airflow-config/generic-web-api/{ENV}-civi-export-processing-state.json'
+    urlSourceType:
+      name: 'civi'
+      typeSpecificVal:
```
de-code: minor: `sourceTypeSpecificValues` might be clearer (otherwise by `type` I would think of basic types such as `string`)

sample_data_config/web-api/web-api-data-pipeline.config.yaml
```diff
@@ -20,16 +20,63 @@ webApi:
...
         defaultPageSize: 100
         pageParameterName: page
     response:
       itemsKeyFromResponseRoot:
         - items
+  # civi export
+  - dataset: '{ENV}'
+    table: civi_data_export
+    stateFile:
+      bucketName: '{ENV}-elife-data-pipeline'
```
de-code, line 37: minor inconsistency with hypothesis config below which uses `ci-elife-data-pipeline`. I understand this is just sample config. I guess in general the configuration we use in production should match, apart from bucket / object in this case.

sample_data_config/web-api/web-api-data-pipeline.config.yaml
tayowonibi: non-intenational

sample_data_config/web-api/web-api-data-pipeline.config.yaml
tayowonibi: acually, the natural form is comma separated values as this is how it is presented to the api, however, i decided to use just a list here for clarity in presentation. So, if you think this should be entered as "csv",  ur opionion will definitinely sway my decision

sample_data_config/web-api/web-api-data-pipeline.config.yaml
tayowonibi: `urlSourceType` instead of `type`?

de-code:
It would be good to have some form of documentation of what the expected response looks like. This is good while the API is not public but also to have it as a reference in the future, should the API change. A link to the API reference might be good: This is what I am looking at: https://docs.civicrm.org/dev/en/latest/api/interfaces/
(It mentions it can return JSON but doesn't provide a sample)

The fields do not seem to match. Maybe it's configuration, version or data specific. e.g. `tag` is not in there.

Maybe this will be part of the test?

sample_data_config/web-api/web-api-data-pipeline.config.yaml
de-code: Just looking at the configuration I am not sure what this is doing or why we need to configure it.
Do you mean all of the values are going to end up comma separated? Should in this case `email` and `source` appear as separate fields?
(I don't have any opinion either way yet)

de-code: I think I only realise now that these are request arguments.

tayowonibi:
> It would be good to have some form of documentation of what the expected response looks like. This is good while the API is not public but also to have it as a reference in the future, should the API change. A link to the API reference might be good: This is what I am looking at: https://docs.civicrm.org/dev/en/latest/api/interfaces/
> (It mentions it can return JSON but doesn't provide a sample)
> 
> The fields do not seem to match. Maybe it's configuration, version or data specific. e.g. `tag` is not in there.
> 
> Maybe this will be part of the test?


Documentation is really poor.
If you have access to our crm site, you can explore the api here https://crm.elifesciences.org/crm/civicrm/api3#explorer.
However, if you don't, you can use their demo site (https://dmaster.demo.civicrm.org/) 
and explore the data returned in the api explorer https://dmaster.demo.civicrm.org/civicrm/api3#explorer.

BTW, you can retrieve user-defined-fields in your civi crm system.

sample_data_config/web-api/web-api-data-pipeline.config.yaml
tayowonibi: > Just looking at the configuration I am not sure what this is doing or why we need to configure it.
> Do you mean all of the values are going to end up comma separated? Should in this case `email` and `source` appear as separate fields?
> (I don't have any opinion either way yet)

`- email,source` is a typo but it won't have which does not have bug-like effect on the data pipeline as I have to concatenate the list as comma seperated values in the web servcice request argument

tayowonibi:
> (It mentions it can return JSON but doesn't provide a sample)

Depending on the request, JSON response can be like this
```
{

    "is_error": 0,
    "version": 3,
    "count": 2,
    "values": [
        {
            "contact_id": "1",
            "last_name": "",
            "contact_is_deleted": "0",
            "id": "1"
        },
        {
            "contact_id": "2",
            "last_name": "Nott",
            "contact_is_deleted": "0",
            "id": "2"
        }
    ]
}
```
OR 
```
{

    "is_error": 0,
    "version": 3,
    "count": 2,
    "values": {
        "1": {
            "contact_id": "1",
            "last_name": "",
            "contact_is_deleted": "0",
            "id": "1"
        },
        "2": {
            "contact_id": "2",
            "last_name": "Nott",
            "contact_is_deleted": "0",
            "id": "2"
        }
    }
}
```

sample_data_config/web-api/web-api-data-pipeline.config.yaml
```diff
@@ -20,21 +20,69 @@ webApi:
...
+      urlExcludingConfigurableParameters: https://crm.elifesciences.org/crm/sites/all/modules/civicrm/extern/rest.php?entity=Contact&action=get
+      configurableParameters:
+        fromDateParameterName: modified_date
+        defaultStartDate: "2012-01-01 00:00:00"
+        dateFormat: "%Y-%m-%d %H:%M:%S"
+        pageSizeParameterName: limit
+        defaultPageSize: 10000
+        offsetParameterName: offset
+      parametersFromEnv:
+        - parameterName: api_key
```
de-code, line 70: I am surprised the main way of passing the key seem to be as a URL parameter. It will be in the access logs can more easily leak.

data_pipeline/generic_web_api/web_api_auth.py
```diff
@@ -0,0 +1,39 @@
...
+        or
+        read_file_content(
+            auth_val_conf.get("valueFileLocation")
+        )
+    )
+
+    return val
+
+
+def read_file_content(file_location: str):
```
de-code: maybe could move this to `utils.pipeline_file_io`

data_pipeline/generic_web_api/helper.py
```diff
@@ -0,0 +1,181 @@
...
+            ] if key and value)
+        param_dict = {
+            **param_dict,
+            **self.compose_able_url_key_val
+        }
+
+        return self.compose_url(param_dict)
+
+
+def get_url_manager(url_source_type: str = None):
```
de-code: would suggest to rename to `get_url_manager_class`, `_factory` or similar.

Maybe `url manager` could also be renamed. The main or only function seem to be to build a url?

data_pipeline/generic_web_api/helper.py
```diff
@@ -0,0 +1,181 @@
+import os
```
de-code: maybe rename the module? this is all url related, correct?

sample_data_config/web-api/web-api-data-pipeline.config.yaml
tayowonibi: you are right

data_pipeline/generic_web_api/helper.py
```diff
@@ -0,0 +1,181 @@
...
+        )
+        options = dict((key, value) for key, value in [
+            (self.offset_param, url_compose_param.page_offset),
+            (
+                self.page_size_param,
+                url_compose_param.page_size or self.page_size
+            )
+            ] if key and value)
+        start_date_param = {
+            self.from_date_param: {">=": start_date}
```
de-code: this seems to end up as a nested dict. e.g.

```python
{
  'modified_date': {
    '>=': '2020-01-01'
  }
}
```

I believe that will be equivalent to (using the string representation of the value / dict):

```python
{
  'modified_date': "{'>=': '2020-01-01'}"
}
```

If that is correct, this should probably explicitly formatted.

de-code: Apologies, I see now that this is formatted using `json.dumps` and passed as the `json` parameter. Maybe that could be reflected in the variable name (`url_query_param` would suggest to me regular url query parameters)

tayowonibi: nested dict is right
check out the api call below
```
https://crm.elifesciences.org/crm/sites/all/modules/civicrm/extern/rest.php?entity=Contact&action=get&api_key=userkey&key=sitekey&json={"sequential":1,"modified_date":{">=":"some_date"},"options":{"limit":100,"offset":2000}}
```

de-code: Is there any documentation that describes that?

de-code:
If we are using the offset, is there a mechanism that ensures that we are not missing out data if data has changed while we are iterating through the pages? What is the result sorted by?

data_pipeline/generic_web_api/helper.py
tayowonibi: > Is there any documentation that describes that?

as far as I know, there is no documentation that describe that, one needs to figure it out by exploring the api using api explorer page of your civi crm deployment

tayowonibi:
> If we are using the offset, is there a mechanism that ensures that we are not missing out data if data has changed while we are iterating through the pages? What is the result sorted by?

it seems to be sorted by `id` by default, i can add a sort by `modified timestamp` to it by default.

data_pipeline/generic_web_api/helper.py
de-code: > > Is there any documentation that describes that?
> 
> as far as I know, there is no documentation that describe that, one needs to figure it out by exploring the api using api explorer page of your civi crm deployment

Okay, I see now. The explorer page is helpful. I can see that this is replacing documentation.
I think I will need to get access to our Civi CRM instance. I can't remember whether I was already given access. I can't find anything. Who should I contact, Rowena?

I might also need access to a web key so that I can at least reproduce the pipeline.

tayowonibi: Contact Rowena for access to our CiviCRM instance.
There is data hub user which Rowena has givne access to read the contacts data via api, and that is the only user whose `api_key`  we can use. Similarly, I have the site key as well and i can share both with you via slack

de-code:
> > If we are using the offset, is there a mechanism that ensures that we are not missing out data if data has changed while we are iterating through the pages? What is the result sorted by?
> 
> it seems to be sorted by `id` by default, i can add a sort by `modified timestamp` to it by default.

Just to see whether my logic is correct (please do correct me in the likely case that this doesn't make sense)...

Sorting by `id` could cause a problem if a record with an earlier id was modified. It could then placed on say the first page which we could have already modified. In that case we would miss out the update of that record.

To sort by `modified_timestamp` might make sense (assuming there are no situations where many records get the same timestamp). If then an entry was modified it would get an updated modified timestamp and should move towards the end. Which might mean that we process records multiple times but do not miss anything out. If many records have the same modified timestamp (perhaps bulk update), then this could create an issue if the order is random. The explorer let's you add multiple `sort` options but may need to check that it would actually use both. In that case it might be good to use the `id` as a secondary sort (might likely need to appear first - if that is supported).

tayowonibi:
> let's you add multiple `sort` options

Only one sort field  is allowed in the resulting url args. i guess ability to add multiple 'sort' fields in the explorer is a bug

data_pipeline/generic_web_api/generic_web_api_config.py
```diff
@@ -145,105 +183,11 @@ def __init__(
...
-            {
-                key: value
-                for key, value in param_dict.items() if key and value
-            }
+class ResponsePathKey:
+    def __init__(self, path_level: str):
+        self.key = path_level if isinstance(path_level, str) else None
+        self.is_variable = (
+            True if isinstance(path_level, dict) and
+            path_level.get('isVariable')
```
de-code: The sample config doesn't seem to include any not-commented out config that is making use of `isVariable`. Maybe we can remove it?

(If we are starting to use variables - maybe it's time to give it the configuration syntax a name, e.g. MCL - Michael Configuration Language :-) )

data_pipeline/generic_web_api/generic_web_api_config.py
```diff
@@ -99,36 +118,55 @@ def __init__(
         next_page_cursor = configurable_parameters.get(
             "nextPageCursorParameterName", None
         )
-        self.url_manager = DynamicURLManager(
+        type_specific_param = api_config.get(
+            "urlSourceType", {}
+        ).get(
+            'sourceTypeSpecificValues', {}
+        )
+        dynamic_url_builder = get_url_builder_class(
```
de-code: would change variable name here as well, e.g. `url_builder_class`

data_pipeline/generic_web_api/generic_web_api_config.py
```diff
@@ -84,9 +85,27 @@ def __init__(
...
         page_size_param = configurable_parameters.get(
             "pageSizeParameterName", None
         )
+        result_sort_param = configurable_parameters.get(
+            "resultSortParameterName", None
+        )
+        result_sort_param_value = configurable_parameters.get(
+            "resultSortParameterValue", None
+        )
+        compose_able_static_parameters = (
```
de-code: The name looks slightly odd because `composeable` is one word.

To me it seems a bit hard to follow maybe because `compose` is a bit generic. I can work it out though. But I think if I had to look at it again in two months I'd probably scratch my head again. I guess those parameters are the ones that end up as URL query parameters.

sample_data_config/web-api/web-api-data-pipeline.config.yaml
```diff
@@ -20,21 +20,71 @@ webApi:
...
+          - tag
+          - group
+          - contact_type
+          - contact_sub_type
+          - last_name
+          - first_name
+          - middle_name
+          - email
+          - source
+          - user_unique_id
```
de-code, line 53: There doesn't seem to be any entry where `user_unique_id` is populated (the `contact_id` and `id` field, which seem to be identical, are returned even if not listed)

sample_data_config/web-api/web-api-data-pipeline.config.yaml
```diff
@@ -20,21 +20,71 @@ webApi:
...
+          - organization_name
+          - phone
+          - city
+          - modified_date
```
de-code, line 57: Just making a note that `modified_date` is returned even if not specified.

sample_data_config/web-api/web-api-data-pipeline.config.yaml
```diff
@@ -20,21 +20,71 @@ webApi:
...
+  # civi export
+  - dataset: '{ENV}'
+    table: civi_data_export
+    stateFile:
+      bucketName: '{ENV}-elife-data-pipeline'
+      objectName: 'airflow-config/generic-web-api/{ENV}-civi-export-processing-state.json'
+    urlSourceType:
+      name: 'civi'
+      sourceTypeSpecificValues:
+        fieldsToReturn:
```
de-code, line 42: Is your suggestion to limit the fields to our use-case and extend later rather than try to just import everything, which may cause issues with GDPR?

sample_data_config/web-api/web-api-data-pipeline.config.yaml
tayowonibi: decision is not implementation decions, but a pipeline configuration decision, and yes, this should be there in the config

sample_data_config/web-api/web-api-data-pipeline.config.yaml
tayowonibi: i have a different opinon about this though, one should specify it if one think it is useful, besides, i am not entirely certain this field is present for all entity types in civi. I'll implement your suggestion though

de-code:
Maybe it's best to move comments regarding the fields at the top level as they remain relevant...

Reviewing the fields you had listed in the your sample config (to also better understand what they mean). I am using the explorer to do that.

| name | example | comment
| ------- | ------------ | -------------
| `external_identifier` | `167045` | this appears to be the eJP person id that we provide
| `group` | `53,310` or `,134,873,1270,317,1342,1357` | The returned field is actually `groups` and it appears to be a comma separated list of numeric ids. Will we make a separate request to find out what they mean or is that for later? (Also found an entry where this list starts with a comma, contact id: `197989`)
| `tag` | `opened vor 29/03/16,opened vor 8/03/16,opened vor 9/02/16,MSA_Cell_biology,MSA_Neuroscience,2018_Feb_Rev_subdrive,Reviewer Survey Q4 2018,Reviewer Survey Q1 2020` | The returned field is called `tags`. Using `tag` as a filter seem to cause issues.
| `source` | `Content alert opt-in form` |
| `organization_name` | | Couldn't spot a populated field. But the filter seem to ignore instructions to find non-empty entries. Maybe because of the type of the field?

Fields that seem to be always empty:

- `user_unique_id`
- `contact_sub_type`


Fields that are always returned:

- `contact_id`
- `id`
- `contact_is_deleted`
- The fields used as a filter (that is why `modified_date` was included)

sample_data_config/web-api/web-api-data-pipeline.config.yaml
tayowonibi: we can import only the fields we need right now like you suggested and extend it later on. To be GDPR compliant if we import some PII in the  civi export table, we can run another pipeline which cleans up the table periodically.

sample_data_config/web-api/web-api-data-pipeline.config.yaml
de-code: This was just a comment / observation on my side.

tayowonibi (commented):
> Reviewing the fields you had listed in the your sample config (to also better understand what they mean). I am using the explorer to do that.

These decision will be made when writing the configuration for the pipeline, and it will be driven by the use case for the external comms

sample_data_config/web-api/web-api-data-pipeline.config.yaml
de-code: I am looking at the fields to understand what they mean and what example values might mean. If we don't have a single example, then I think we can't be sure it's actually working or would work. (I moved this to a top level PR comment as it will be good to have a list of example values for some of the fields)

de-code:
> > Reviewing the fields you had listed in the your sample config (to also better understand what they mean). I am using the explorer to do that.
> 
> These decision will be made when writing the configuration for the pipeline, and it will be driven by the use case for the external comms

Okay, fair enough.

For some values, like the groups this could mean that we will need code changes. Although I guess we could use the same code but just change the URL to use another entity, groups in that case.

Not sure what the best place for this list is. But it would be good to document some of the fields somewhere.

sample_data_config/web-api/web-api-data-pipeline.config.yaml
de-code: Okay, just good to also understand the motivation behind configuring the list of fields.

Would be good to get more thoughts from you on the cleanup you mentioned.
Would we the reset the start modified date state, re-import everything and then delete anything we imported before starting that. Although now that I typed that, I guess we would just import the `contact_is_deleted` flag and delete anything at our end where this is set?

tayowonibi: run a pipeline that deletes row with id's  whose latest record's `contact_is_deleted` is set to true

de-code:
Are you still planning to add more tests to this PR?

How did you develop locally? Did you say that you used a local Civi instance?

tayowonibi:
> For some values, like the groups this could mean that we will need code changes. Although I guess we could use the same code but just change the URL to use another entity, groups in that case.

In this case, I'll have a new DAG dedicated to civi import with its own dedicated config separated from the generic web data pipeline. In this I'll import all necessary entities, and have a function run a query that links all entities together

de-code:
> > For some values, like the groups this could mean that we will need code changes. Although I guess we could use the same code but just change the URL to use another entity, groups in that case.
> 
> In this case, I'll have a new DAG dedicated to civi import with its own dedicated config separated from the generic web data pipeline. In this I'll import all necessary entities, and have a function run a query that links all entities together

Might be good to have a chat about that in our catch-up.

tayowonibi:
@de-code - i think that is all the modifications I need to make to the repo for now.

sample_data_config/web-api/web-api-data-pipeline.config.yaml
de-code: Maybe revert this back to profile?

tests/unit_test/test_generic_web_api_data_etl.py
```diff
@@ -0,0 +1,293 @@
...
+    def test_should_return_all_data_when_data_is_a_list(
+            self
+    ):
+        data_config = get_data_config(WEB_API_CONFIG)
+        data = [['first'], ['second'], ['third']]
+        expected_time_list = get_items_list(
+            data,
+            data_config,
+        )
+        assert data == expected_time_list
```
de-code: is this not the reverse, `data` here is the expected items_list and `expected_time_list` is the actual value? (also true for the following test)

(also good to keep the order actual == expected for consistency, some of the tests here have the reverse order)

tests/unit_test/test_generic_web_api_data_etl.py
```diff
@@ -0,0 +1,293 @@
...
+            data_config,
+        )
+        assert expected_response == actual_response
+
+
+class TestNextCursor:
+    def test_should_be_none_when_cursor_parameter_is_not_in_config(self):
+        data_config = get_data_config(WEB_API_CONFIG)
+        data = {'key': 'val', 'values': []}
+        assert not get_next_cursor_from_data(data, data_config)
```
de-code: maybe check `is None` if that is what it is meant to return?

de-code (approved):

PR #97, opened 2020-07-14
bug fix and modify for toggl data download requirements

tayowonibi:


data_pipeline/generic_web_api/url_builder.py
```diff
@@ -51,6 +51,7 @@ def __init__(
             compose_able_url_key_val: dict = None,
             sort_key: str = None,
             sort_key_val: str = None,
+            default_start_end_date_diff_in_days: int = None,
```
de-code: `default_start_end_date_diff_in_days` doesn't seem to be used in this class.

tayowonibi: fixed

data_pipeline/generic_web_api/generic_web_api_data_etl.py
```diff
@@ -157,18 +160,27 @@ def generic_web_api_data_etl(
             )
             items_count = len(items_list)
 
+            from_date_to_advance, to_reset_other_request_param = (
```
de-code, line 163: It would be good to have tests that cover this functionality.

I found the naming a bit confusing at first. Maybe `should_reset_pagination` would be slightly clearer. That should be true whenever the start date has changed. Maybe moving that logic out of `get_next_start_date` could make it a bit more obvious.

data_pipeline/generic_web_api/generic_web_api_config.py
```diff
@@ -82,6 +82,11 @@ def __init__(
         ).get("configurableParameters", {})
         self.default_start_date = configurable_parameters.get(
             "defaultStartDate", None)
+        default_start_end_date_diff_in_days = (
+            configurable_parameters.get(
+                "defaultDaysDiffFromStartTillEnd", None
```
de-code: minor: why is it called default? It seems to be the only configuration value we use.

minor: usually we put the type / unit at the end, like in the variable name. I was trying to think what name I personally would think of initially. Maybe something like `queryPeriodInDays`. Although your name refers to the affected parameters and may be better that way.

docker-compose.yml
```diff
@@ -23,6 +23,8 @@ x-airflow-env:
     - GOOGLE_ANALYTICS_CONFIG_FILE_PATH=/usr/local/airflow/app-config/google-analytics/ga-data-pipeline.config.yaml
     - CIVICRM_API_KEY=SOME_API_KEY
     - CIVICRM_SITE_KEY=SOME_SITEKEY
+    - PATH_TO_TOGGL_API_TOKEN_FILE=SOME_TOGGLE_API_TOKEN_FILE_PATH
```
de-code: minor: I am not sure whether I had added a comment somewhere else. For consistency I would suggest something like `TOGGLE_API_TOKEN_FILE_PATH` for the env variable name

de-code (approved):

PR #109, opened 2020-07-26
modification made for paged import of civi data

tayowonibi:
part of https://github.com/elifesciences/issues/issues/5633

### DIFF
--- a/data_pipeline/generic_web_api/generic_web_api_data_etl.py
+++ b/data_pipeline/generic_web_api/generic_web_api_data_etl.py
@@ -6,7 +6,6 @@ from pathlib import Path
 import json
 from json.decoder import JSONDecodeError
 
-from google.cloud.bigquery import WriteDisposition
 from botocore.exceptions import ClientError
 
 from data_pipeline.generic_web_api.transform_data import (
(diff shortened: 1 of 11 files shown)


docker-compose.dev.override.yml
```diff
@@ -1,23 +1,36 @@
...
+  &airflow-env
+    - CIVICRM_API_KEY_FILE_PATH=/usr/local/airflow/secret/civi_api_key
+    - CIVICRM_SITE_KEY_FILE_PATH=/usr/local/airflow/secret/civi_site_key
+    - PATH_TO_TOGGL_API_TOKEN_FILE=/usr/local/airflow/toggl.txt
+
 
 x-airflow-volumes:
   &airflow-volumes
     - ~/.config/gcloud/application_default_credentials.json:/tmp/credentials.json
+    - ./civi_site_key:/usr/local/airflow/secret/civi_site_key
```
de-code: minor: we probably need some developer documentation relating to those keys. The gcloud and AWS credentials are pretty standard but these are not.
I suspect if they don't exist, then Docker will create directories with those names which can be confusing. Maybe if they live locally, it could be easier to mount a `.secrets` directory with those three files in it.
What do you think?

tayowonibi: moved them all to a directory

data_pipeline/generic_web_api/generic_web_api_data_etl.py
```diff
@@ -205,13 +223,19 @@ def load_written_data_to_bq(
             full_temp_file_location,
             quoted_values_are_strings=False
         )
+        write_disposition = (
```
de-code: very minor: I'd probably move that to the configuration loading

de-code (approved):

PR #351, opened 2021-02-10
Crossref eventdata failure

HazalCiplak:


data_pipeline/crossref_event_data/etl_crossref_event_data_util.py
```diff
@@ -220,6 +224,13 @@ def per_doi_download_page_etl(
         schema: list,
         journal_previous_timestamp: datetime,
 ):
+    http.client.HTTPConnection.debuglevel = 1
```
de-code: Is this to get the output in the error case?

Enabling debug logging at the lower level might log things we may not want to log (e.g. credentials).
(It also affects not just this DAG I think)

I would suggest to just get it from the response (I guess inside `get_crossref_data_single_page`).
We should have the full response with the status code available.

HazalCiplak: OK I am checking thank you!

sample_data_config/crossref-event/crossref-event-data-pipeline.config.yaml
```diff
@@ -18,5 +18,5 @@ schemaFile:
 # the bigquery data warehouse
 publisherIdPrefixes:
   - '10.7554'
-crossrefEventBaseUrl: "https://api.eventdata.crossref.org/v1/events?rows=1000"
+crossrefEventBaseUrl: 'https://api.eventdata.crossref.org/v1/events?rows=100&mailto=h.ciplak@elifesciences.org'
```
de-code, line 21: just mentioning that this repo is public, not sure if email crawlers will find it (and if you mind it). But maybe not.

de-code (approved):

PR #391, opened 2021-03-11
Adding gmail data to the pipeline

HazalCiplak:
https://github.com/elifesciences/data-hub-issues/issues/144


data_pipeline/gmail_production_data/get_gmail_data.py
```diff
@@ -0,0 +1,114 @@
...
+    return:
+        - df : DataFrame
+    '''
+    message_list = []
+    pages = 0
+    limit = 10000
+    # get emails for specific labels
+    for labelId in labels_need:
+        pages = 0
+        response = service.users().messages().list(userId=userId, labelIds=labelId).execute()
```
de-code: do you find that the inbox contains many unrelated messages?

data_pipeline/gmail_production_data/get_gmail_data.py
```diff
@@ -0,0 +1,114 @@
...
+            response = service.users().messages().list(
+                userId=userId,
+                labelIds=labelId,
+                pageToken=page_token
+                ).execute()
+            message_list.extend(response['messages'])
+            pages = pages + 1
+
+    # to get unique list of dict list
+    list_of_unique_dicts = list(np.unique(np.array(message_list).astype(str)))
```
de-code: I have to admit that I don't fully understand this section of the code. Perhaps you could explain it to me in our catch-up.

data_pipeline/gmail_production_data/get_gmail_data.py
HazalCiplak: Yes unfortunately there is more than 100 labels so need to filter by them.

data_pipeline/gmail_production_data/get_gmail_data.py
```diff
@@ -0,0 +1,114 @@
...
+
+    credentials = service_account.Credentials.from_service_account_file(
+        SERVICE_ACCOUNT_FILE, scopes=SCOPES)
+    delegated_credentials = credentials.with_subject(EMAIL_ADDRESS)
+
+    service = build("gmail", "v1", credentials=delegated_credentials)
+    return service
+
+
+def get_label_list(service, userId):
```
de-code: I would suggest to make use of Python's typing hints (and we could check them using `mypy`).
Personally, I try to add as little as possible code comments. Some comments may be a considered a "code smell". I can't find a good link but this is one of the results: https://softwareengineering.stackexchange.com/questions/1/comments-are-a-code-smell

Just as an example, not to critique your comment (because the same would be true for my comments sooner or later): the comment is out of sync as it is not writing to a CSV file. The parameter list is incomplete and if we added the type hints would just repeat the same information.

i.e. still add comments that you think are useful but not in the code. But in some cases we could change the code to include that information.

One information that is not be as easy to read from the code, but might be relevant is what fields does this actually return, as it's not just a list of strings.
I guess in this case the fields of https://developers.google.com/gmail/api/reference/rest/v1/users.labels#Label
I suppose the labels().list method already has a reference to it, but I personally would find it useful (although not necessary) here with maybe the main fields "id" and "name.

data_pipeline/gmail_production_data/get_gmail_data.py
```diff
@@ -0,0 +1,114 @@
...
+    param:
+        - service : connection
+        - userId : str
+        - label : list
+    return:
+        - df : DataFrame
+    '''
+    message_list = []
+    pages = 0
+    limit = 10000
```
de-code: do we need the limit?

data_pipeline/gmail_production_data/get_gmail_data.py
```diff
@@ -0,0 +1,114 @@
...
+    # get emails for specific labels
+    for labelId in labels_need:
+        pages = 0
+        response = service.users().messages().list(userId=userId, labelIds=labelId).execute()
+        if 'messages' in response:
+            #  to add the end of the list
+            message_list.extend(response['messages'])
+        while ('nextPageToken' in response) and (pages < limit):
```
de-code: I personally found it helpful to me to separate the paging from the extraction, e.g. it could be a simple python generator function that yields pages of messages, or in some cases individual messages.
In that case this code block would be mostly the code of that generator function, but rather than appending to `message_list` it would have `yield from response['messages']`.

data_pipeline/gmail_production_data/get_gmail_data.py
de-code: If we get all of the message thread id pairs for the production account, what will be the size of the file? i.e. is there a chance we are hitting memory limits?

de-code:
My personal recommendation would be to stick with the simple pipeline and get this through. And add the thread details as a separate PR. Otherwise there will become PR might become too big to review.

data_pipeline/gmail_production_data/get_gmail_data.py
```diff
@@ -0,0 +1,141 @@
...
+
+    credentials = service_account.Credentials.from_service_account_file(
+        SERVICE_ACCOUNT_FILE, scopes=SCOPES)
+    delegated_credentials = credentials.with_subject(EMAIL_ADDRESS)
+
+    service = build("gmail", "v1", credentials=delegated_credentials)
+    return service
+
+
+def retry(ExceptionToCheck, tries=5, delay=4, backoff=3, logger=None):
```
de-code: I would recommend using https://pypi.org/project/backoff/ we have used it in other Python projects as well.

data_pipeline/gmail_production_data/get_gmail_data.py
```diff
@@ -0,0 +1,141 @@
...
+        yield from response['messages']
+
+
+def get_link_message_thread(service, userId):
+    df = pd.DataFrame(iter_link_message_thread(service, userId))
+    return df
+
+
+@retry(Exception, tries=5)
+def get_one_thread(service, userId, threadId):
```
de-code: why don't we just dump the response into BigQuery and do the calculation there?

data_pipeline/gmail_production_data/get_gmail_data.py
```diff
@@ -0,0 +1,141 @@
...
+    return df
+
+
+def get_thread_details_for_given_ids(service, userId, threadIds_file):
+    df = pd.read_csv(threadIds_file, usecols=['threadId'])
+    threadIds_arr = df['threadId'].unique()
+    df = pd.DataFrame()
+    for threadId in threadIds_arr:
+        df_one_thread = get_one_thread(service, userId, threadId)
+        df = df.append(df_one_thread)
```
de-code: [merge](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.merge.html) would probably be more efficient. Perhaps this is something where the data will get very large? Or not so large because you are doing the counting?

EDIT: actually I meant [concat](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.concat.html)

data_pipeline/gmail_production_data/get_gmail_data.py
```diff
@@ -0,0 +1,141 @@
...
+
+    return df
+
+
+def write_dataframe_to_file(df_data_to_write, target_file_path):
+    df_data_to_write.to_csv(target_file_path, encoding='utf-8-sig', index=False)
```
de-code: To be honest I have never used `utf-8-sig` before. What is the benefit of using it over `utf-8`?

data_pipeline/gmail_production_data/get_gmail_data.py
```diff
@@ -0,0 +1,141 @@
...
+        - total_email_count
+        - sent_email_count
+        - first_responder
+        - first_reponse_date
+    """
+    thread_results = service.users().threads().get(userId=userId, id=threadId).execute()
+
+    df = pd.DataFrame()
+
+    df['threadId'] = [thread_results['id']]
```
de-code: Perhaps if there is a single row, then a dict would be slightly easier to follow?

data_pipeline/gmail_production_data/get_gmail_data.py
HazalCiplak: Himm not sure how to use it, `merge` is working like `join` but I need to do here `union`

data_pipeline/gmail_production_data/get_gmail_data.py
HazalCiplak: Lets talk about it at our data meeeting

de-code:
Just to make sure you are not blocked by me: please let me know when you would like me to review more

data_pipeline/gmail_data/get_gmail_data.py
```diff
@@ -0,0 +1,45 @@
+from typing import Iterable
+import pandas as pd
+
+from google.oauth2 import service_account
+from googleapiclient.discovery import build
+from googleapiclient.discovery import Resource
+
+
+# this connection type will be changed !!
```
de-code: perhaps remove this comment as we have the GitHub issue and are not going to change it as part of this PR?

data_pipeline/gmail_data/get_gmail_data.py
```diff
@@ -0,0 +1,45 @@
...
+def connect_to_email(userId: str) -> Resource:
+    SCOPES = ['https://www.googleapis.com/auth/gmail.readonly']
+    SERVICE_ACCOUNT_SECRET_FILE = 'DELETE/elife-data-pipeline-test.json'
```
de-code: perhaps make this configurable, maybe in this case it makes sense if the path to the file is passed in from the DAG?

data_pipeline/gmail_data/get_gmail_data.py
```diff
@@ -0,0 +1,45 @@
...
+    EMAIL_ADDRESS = userId
```
de-code: The separate `email_address` may not add so much. Perhaps remove and just use `userId`? (Or rename the passed in parameter?)

data_pipeline/gmail_data/get_gmail_data.py
```diff
@@ -0,0 +1,45 @@
+from typing import Iterable
+import pandas as pd
+
+from google.oauth2 import service_account
+from googleapiclient.discovery import build
+from googleapiclient.discovery import Resource
+
+
+# this connection type will be changed !!
+def connect_to_email(userId: str) -> Resource:
```
de-code: minor: naming, perhaps rename parameter to `user_id`?

Maybe worth also reviewing the function name. But it's always difficult. Perhaps build gmail service, but then it's also getting credentials. Just a thought anyway.

EDIT: In the DAG it's called `get_gmail_service`, so maybe something like `get_gmail_service_for_user_id`?

data_pipeline/gmail_data/get_gmail_data.py
```diff
@@ -0,0 +1,45 @@
...
+        response = service.users().messages().list(userId=userId, pageToken=page_token).execute()
+        yield from response['messages']
+
+
+def get_link_message_thread(service: Resource, userId: str) -> pd.DataFrame:
+    df = pd.DataFrame(iter_link_message_thread(service, userId))
+    return df
+
+
+def write_dataframe_to_file(df_data_to_write: pd.DataFrame, target_file_path: str):
```
de-code: minor: now that we came across CSV and JSONL files, might be good to add csv to the function name? (e.g. `write_dataframe_to_csv_file`)

dags/gmail_label_data_import_pipeline.py
```diff
@@ -0,0 +1,155 @@
...
+)
+
+from data_pipeline.gmail_data.get_gmail_data import (
+    connect_to_email,
+    get_label_list,
+    write_dataframe_to_file,
+    # get_link_message_thread,
+)
+
+USER_ID_ENV_NAME = "GMAIL_DATA_USER_ID"
```
de-code: minor: just thinking, now that we have a config file, would it make sense to add to the config file instead?

dags/gmail_label_data_import_pipeline.py
```diff
@@ -0,0 +1,155 @@
...
+        filename = Path(tmp_dir)/data_config.table_name_labels
+        write_dataframe_to_file(get_label_list(get_gmail_service(), USER_ID_ENV_NAME), filename)
+        LOGGER.info('Created file: %s', filename)
+        load_file_into_bq(
+            filename=filename,
+            dataset_name=data_config.dataset,
+            table_name=data_config.table_name_labels,
+            project_name=data_config.project_name,
+            source_format=bigquery.SourceFormat.CSV
+            )
```
de-code: minor: indentation

dags/gmail_label_data_import_pipeline.py
```diff
@@ -0,0 +1,155 @@
...
+from data_pipeline.gmail_data.get_gmail_data import (
+    connect_to_email,
+    get_label_list,
+    write_dataframe_to_file,
+    # get_link_message_thread,
+)
+
+USER_ID_ENV_NAME = "GMAIL_DATA_USER_ID"
+GMAIL_DATA_CONFIG_FILE_PATH_ENV_NAME = "GMAIL_DATA_CONFIG_FILE_PATH"
+DEPLOYMENT_ENV_NAME = "DEPLOYMENT_ENV"
```
de-code: I would perhaps keep it `DEPLOYMENT_ENV_ENV_NAME` just for consistency (not an ideal name but it was mean "env name" of the "deployment env").

Just a comment: There seem to be a bit of duplication across the DAGs, like this constant. Worth refactoring. We currently have https://github.com/elifesciences/data-hub-issues/issues/7 which may partially cover it, although in this case its duplicated within the same repo.

dags/gmail_label_data_import_pipeline.py
```diff
@@ -0,0 +1,155 @@
...
+    conf_file_path = get_env_var_or_use_default(
+        GMAIL_DATA_CONFIG_FILE_PATH_ENV_NAME, ""
+    )
+    LOGGER.info('conf_file_path: %s', conf_file_path)
+    data_config_dict = get_yaml_file_as_dict(conf_file_path)
+    LOGGER.info('data_config_dict: %s', data_config_dict)
+    kwargs["ti"].xcom_push(
+        key="data_config_dict",
+        value=data_config_dict
+        )
```
de-code: minor: indentation

dags/gmail_label_data_import_pipeline.py
```diff
@@ -0,0 +1,155 @@
...
+    write_dataframe_to_file,
+    # get_link_message_thread,
+)
+
+USER_ID_ENV_NAME = "GMAIL_DATA_USER_ID"
+GMAIL_DATA_CONFIG_FILE_PATH_ENV_NAME = "GMAIL_DATA_CONFIG_FILE_PATH"
+DEPLOYMENT_ENV_NAME = "DEPLOYMENT_ENV"
+DEFAULT_DEPLOYMENT_ENV = "ci"
+
+DAG_ID = "Get_Gmail_Label_Data"
```
de-code: Is this DAG only going to do label data? I would make this consistent with the filename. That will make it easier to go from DAG in Airflow to the corresponding Python file.

EDIT: We could also just rename the file. I guess it's good to have the name reflect what it is currently doing and worry about the future name when we have future functionality.

dags/gmail_label_data_import_pipeline.py
```diff
@@ -0,0 +1,155 @@
...
+
+from data_pipeline.utils.pipeline_file_io import (
+    get_yaml_file_as_dict
+)
+
+from data_pipeline.gmail_data.get_gmail_data import (
+    connect_to_email,
+    get_label_list,
+    write_dataframe_to_file,
+    # get_link_message_thread,
```
de-code: minor: remove

sample_data_config/gmail-data/data-schema/gmail_label_list_schema.json
```diff
@@ -0,0 +1,32 @@
...
+        "type": "STRING"
+    },
+    {
+        "mode": "NULLABLE",
+        "name": "labelType",
+        "type": "STRING"
+    },
+    {
+        "mode": "NULLABLE",
+        "name": "labelColour",
```
de-code: Are the field ids in here returned by the API? I only noticed it because names are usually americanised. The API docs have slightly different names, e.g. `color`: https://developers.google.com/gmail/api/reference/rest/v1/users.labels#Label (is that the correct API docs?)
(But you used the API and will know better)

data_pipeline/gmail_data/get_gmail_data.py
```diff
@@ -0,0 +1,45 @@
...
+import pandas as pd
+
+from google.oauth2 import service_account
+from googleapiclient.discovery import build
+from googleapiclient.discovery import Resource
+
+
+# this connection type will be changed !!
+def connect_to_email(userId: str) -> Resource:
+    SCOPES = ['https://www.googleapis.com/auth/gmail.readonly']
```
de-code: (may have commented before but can't see it)
I would suggest to move all constants out of the function (may just be this one). In this case maybe also add a prefix, e.g. `GMAIL_SCOPES`

de-code:
I added a few comments, nothing major (please feel free to disagree).

EDIT: by that I mean: it's good, and almost ready to get merged :smile:

Another thing I can think of (I didn't forget yet :wink:), is adding tests. My suggestion (and it is just a suggestion), would be to get into the habit of adding or changing at least one test per PR. From the changes in this PR, I could think that adding a test for the config could be the easiest, as it doesn't require a mock.

data_pipeline/gmail_data/get_gmail_data.py
HazalCiplak: I have changed the function name.

dags/gmail_label_data_import_pipeline.py
HazalCiplak: Your eyes are better than `pylint` :D

data_pipeline/gmail_data/get_gmail_data.py
```diff
@@ -0,0 +1,40 @@
+from typing import Iterable
+import pandas as pd
+
+from google.oauth2 import service_account
+from googleapiclient.discovery import build
+from googleapiclient.discovery import Resource
+
+
+def get_gmail_service_for_user_id(secretFile: str, scopes: str, userId: str) -> Resource:
```
de-code: minor: would change the parameters here and below to `snake_case`

`pylint` would usually pick that up I think but currently we have the `invalid-name` pylint exclusion configured.

dags/gmail_data_import_pipeline.py
```diff
@@ -0,0 +1,170 @@
...
+    LOGGER.info('data_config: %s', data_config)
+    return data_config
+
+
+def get_gmail_service():
+    return get_gmail_service_for_user_id(
+        GMAIL_ACCOUNT_SECRET_FILE,
+        GMAIL_SCOPES,
+        GMAIL_DATA_USER_ID_ENV
+        )
```
de-code: minor: indentation

dags/gmail_data_import_pipeline.py
```diff
@@ -0,0 +1,170 @@
...
+)
+
+from data_pipeline.gmail_data.get_gmail_data import (
+    get_gmail_service_for_user_id,
+    get_label_list,
+    write_dataframe_to_csv_file
+)
+
+GMAIL_SCOPES = ['https://www.googleapis.com/auth/gmail.readonly']
+GMAIL_ACCOUNT_SECRET_FILE = 'DELETE/elife-data-pipeline-test.json'
```
de-code: This still has the `DELETE` (and `test`) in the path. Perhaps make it configurable? (apologies, if that was still in progress)

dags/gmail_data_import_pipeline.py
```diff
@@ -0,0 +1,170 @@
...
+
+GMAIL_DATA_USER_ID_ENV = "GMAIL_DATA_USER_ID"
+GMAIL_DATA_CONFIG_FILE_PATH_ENV_NAME = "GMAIL_DATA_CONFIG_FILE_PATH"
+DEPLOYMENT_ENV_ENV_NAME = "DEPLOYMENT_ENV"
+DEFAULT_DEPLOYMENT_ENV = "ci"
+
+DAG_ID = "Gmail_Data_Pipeline"
```
de-code: minor: thank you for changing. Just noticed still not completely matching the filename, with the current filename it would be `Gmail_Data_Import_Pipeline`.

HazalCiplak: I think the file is for importing the job to pipeline but the dag name is pipeline itself ?

de-code: I would agree to some degree. But I would argue that the purpose of the file is to define the DAG (as it is in the `dags` directory), as such it seems to make sense for them to be consistent. (I think I may have read something about best practices but can't find it now)

The other argument would be that it would be consistent with what we do in other DAGs, I think.

Whether we should have `pipeline` in the filename and DAG ID is another question. We have for example controller DAGs that don't.

HazalCiplak: OK I see, I am updating

sample_data_config/gmail-data/data-schema/gmail_label_list_schema.json
de-code: I think I am still confused how the import with CSV works. We don't seem to be providing the schema to the bigquery table. It seems odd if the bigquery import itself was to assume that the columns in the CSV match exactly the column order of the BigQuery table (but everything is possible). If that was really the case, then we could probably just select the columns in the right order you don't want to switch to JSONL.

Another thing I haven't thought of before is that the default mode is to append. i.e. every time you run the pipeline, it should append updated labels. If we keep that append mode, then we need at least an `imported_timestamp`.

dags/gmail_data_import_pipeline.py
```diff
@@ -0,0 +1,170 @@
...
+        json_schema=schema_json
+    )
+    LOGGER.info('Created table: %s', data_config.table_name_labels)
+
+
+def gmail_label_data_etl(**kwargs):
+    data_config = data_config_from_xcom(kwargs)
+    user_id = get_env_var_or_use_default(GMAIL_DATA_USER_ID_ENV, "")
+    with TemporaryDirectory() as tmp_dir:
+        filename = Path(tmp_dir)/data_config.table_name_labels
```
de-code: minor: would suggest to add spaces around the `/` operator

de-code:
Just noting down here what we need to do for the deployment (we can still discuss it):

Locally you are using the service account credentials file that is mounted into the container.
When deployed in staging and production, we will also mount a file into the container.
But that file will come from Kubernetes secrets.
When the Data Hub formula runs, it is retrieving secrets from `vault` and putting them into Kubernetes secrets.

i.e. we will need:

- add service account credentials to vault (Joel will do that)
- change the `Jenkinsfile` of the Data Hub formula to add the new secret (should just be one line)
- change the Data Hub formula application config to mount the secret to the path specified by the environment variable

dags/gmail_data_import_pipeline.py
```diff
@@ -0,0 +1,155 @@
...
+
+        create_table_if_not_exist(
+            project_name=data_config.project_name,
+            dataset_name=data_config.dataset,
+            table_name=data_config.table_name_labels,
+            json_schema=generated_schema
+        )
+        LOGGER.info('Created table: %s', data_config.table_name_labels)
+
+        if filename.lower().endswith('.csv'):
```
de-code: Currently `generate_schema_from_file` (called above) doesn't seem to support CSV files. I would maybe remove the CSV handling as it would probably fail now?

data_pipeline/gmail_data/get_gmail_data.py
```diff
@@ -0,0 +1,63 @@
...
+    return service
+
+
+@backoff.on_exception(backoff.expo, TimeoutError, max_tries=10)
+def get_label_list(service: Resource, user_id: str) -> pd.DataFrame:
+    imported_timestamp = datetime.datetime.utcnow().isoformat()
+    label_response = service.users().labels().list(userId=user_id).execute()
+    df_one_label = pd.DataFrame()
+    df_label = pd.DataFrame()
+    for i in range(len(label_response['labels'])):
```
de-code: minor: I would write this as:

```python
for label in label_response['labels']:
```

de-code:
I noticed that `pylint` is currently not running over the `data_pipeline.gmail_data` package because there is no `__init__.py`. Once added, it will have a few small complaints.

It will also complain about `df` which we can add to `good-names` in `.pylintrc` by adding (see [pylint docs](https://pylint.pycqa.org/en/latest/technical_reference/features.html#basic-checker-options)):

```
[FORMAT]
good-names=i,j,k,ex,Run,_,df
```

data_pipeline/gmail_data/get_gmail_data.py
```diff
@@ -0,0 +1,63 @@
...
+        response = service.users().messages().list(userId=user_id, pageToken=page_token).execute()
+        yield from response['messages']
+
+
+def get_link_message_thread(service: Resource, user_id: str) -> pd.DataFrame:
+    df = pd.DataFrame(iter_link_message_thread(service, user_id))
+    return df
+
+
+def write_dataframe_to_file(
```
de-code: minor: I would maybe suggest to remove the CSV functionality unless we really want to use it. If we then rename it to say `write_dataframe_to_jsonl_file` then it would be quite clear to the caller what file formats it supports.
We have a similar function for regular list of dicts in `data_pipeline/utils/pipeline_file_io.py`, it might be an idea to move it there?

de-code (approved):

docker-compose.override.yml
```diff
@@ -5,6 +5,7 @@ x-airflow-env:
     - CIVICRM_API_KEY_FILE_PATH=/home/airflow/secrets/civi_api_key
     - CIVICRM_SITE_KEY_FILE_PATH=/home/airflow/secrets/civi_site_key
     - PATH_TO_TOGGL_API_TOKEN_FILE=/home/airflow/secrets/toggl.txt
+    - GOOGLE_APPLICATION_CREDENTIALS=/home/airflow/secrets/gmail_credentials.json
```
de-code, line 8: Just a note: This will override the "default credentials", that are mounted below from `~/.config/gcloud/application_default_credentials.json` to `/tmp/credentials.json`, and then copied to `/home/airflow/.config/gcloud/` via the command defined in `docker-compose.yml` file.
i.e. in development I guess all GCP access will only work with the gmail service account credentials file.

sample_data_config/gmail-data/data-schema/gmail_label_list_schema.json
```diff
@@ -0,0 +1,32 @@
+[
```
de-code: is the schema file still used?

dags/gmail_data_import_pipeline.py
```diff
@@ -0,0 +1,155 @@
...
+        filename = os.path.join(tmp_dir, data_config.stage_file_name_labels)
+
+        write_dataframe_to_file(
+            df_data_to_write=get_label_list(get_gmail_service(),  user_id),
+            target_file_path=filename
+        )
+
+        LOGGER.info('Created file: %s', filename)
+
+        generated_schema = generate_schema_from_file(filename)
```
de-code, line 100: minor: you could use `create_or_extend_table_schema`, I believe that also calls the `generate_schema_from_file`.

sample_data_config/gmail-data/data-schema/gmail_label_list_schema.json
HazalCiplak: removed

data_pipeline/gmail_data/get_gmail_data.py
HazalCiplak: It is noted. In case of write to google cloud storage I will keep it here and later I can add it when I deploy all.

de-code: You could still use JSON when writing to google cloud storage as well.

EDIT: Apologies, I realise my comment doesn't make much sense. You meant keeping it in the module.

dags/gmail_data_import_pipeline.py
HazalCiplak: This function name does not look explicit it is like about creating schema but not about creating a table, so did not want to use it. Maybe later I can change the name of the function to use it.

data_pipeline/gmail_data/get_gmail_data.py
HazalCiplak: Yes for not to work on different files Just want to keep it here for now. Then I can move it.

PR #405, opened 2021-03-22
Add gmail thread details

HazalCiplak:
https://github.com/elifesciences/data-hub-issues/issues/158


data_pipeline/gmail_data/get_gmail_data.py
```diff
@@ -46,14 +52,94 @@ def iter_link_message_thread(service: Resource, user_id: str) -> Iterable[dict]:
...
-    df_link = pd.DataFrame(iter_link_message_thread(service, user_id))
+    imported_timestamp = get_current_timestamp()
+    df_link = pd.DataFrame(iter_link_message_thread_ids(service, user_id))
+    df_link['user_id'] = user_id
+    df_link['imported_timestamp'] = imported_timestamp
     return df_link
 
 
+@backoff.on_exception(backoff.expo, TimeoutError, max_tries=10)
+def get_one_thread(service: str, user_id: str, thread_id: str) -> pd.DataFrame:
```
de-code, line 63: Here I would still wonder whether it would be beneficial to just dump the data into BigQuery and do the counts via BigQuery views or Data Studio reports? (see https://github.com/elifesciences/data-hub-core-airflow-dags/pull/391#discussion_r592527299)

data_pipeline/gmail_data/get_gmail_data.py
```diff
@@ -46,14 +52,94 @@ def iter_link_message_thread(service: Resource, user_id: str) -> Iterable[dict]:
...
+            if header['name'] == 'From':
+                if email_sent_pattern in header['value']:
+                    sent_email_count += 1
+
+    df_thread['sent_email_count'] = sent_email_count
+
+    # to get first responder
+    # pylint: disable=consider-using-enumerate
+    first_response_index = 0
+    for i in range(len(messages)):
```
de-code: BTW if you need to know the index while iterating then you can use `enumerate`:

```python
for i, message in enumerate(messages):
```

data_pipeline/gmail_data/get_gmail_data.py
```diff
@@ -46,14 +52,94 @@ def iter_link_message_thread(service: Resource, user_id: str) -> Iterable[dict]:
...
+        SELECT DISTINCT {}
+        FROM  `{}.{}.{}`
+        """.format(column_name, project_name, dataset, table_name)
+    )
+
+    client = bigquery.Client(project_name)
+    query_job = client.query(sql)  # API request
+    results = query_job.result()  # Waits for query to finish
+
+    df_value = pd.DataFrame()
```
de-code: I am not 100% sure what this is doing. Is it generating a DataFrame with a single column (`column_name`) where the value comes from `row.threadId`?

If so, this would perhaps be shorter?

```python
df_value = pd.DataFrame({
    {content_name: row.threadId}
} for row in results)
```

But you could also consder using pandas' [read_gbq](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_gbq.html). (I am using it in [data-science-dags](https://github.com/elifesciences/data-science-dags/blob/46559fa41d8641137981d1490466e4e7363d0a3a/data_science_pipeline/utils/jupyter.py#L18), but that probably doesn't help much because it's called by notebooks).

---

As a side note, when I suggested [pd.concat](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.concat.html), I think it will only really make things faster if you call it once (not per row). e.g.:

```python
pd.concat([
    some_method_returning_a_dataframe(item)
    for item in items
])
```

e.g. if `items` was a list of thread ids and `some_method_returning_a_dataframe` was a function that returned a DataFrame containing messages for that thread id.

dags/gmail_data_import_pipeline.py
```diff
@@ -148,7 +150,63 @@ def gmail_link_message_thread_ids_etl(**kwargs):
...
+    df_thread_details = pd.DataFrame()
+    df_thread_details_temp = pd.DataFrame()
+
+    for _, row in df_thread_id_list.iterrows():
+        df_thread_details_temp = get_one_thread(
+                                    get_gmail_service(),
+                                    user_id,
+                                    row[data_config.column_name_list_of_thread_ids]
+                                )
+        df_thread_details = pd.concat(
```
de-code: I see this now, this seems to be a case where `pd.concat` can be benefical, but would need to be outside the loop. e.g.

```python
service = get_gmail_service()
thread_ids = df_thread_id_list[data_config.column_name_list_of_thread_ids]
pd.concat([
    get_one_thread(service, user_id, thread_id)
    for thread_id in thread_ids
])
```

(if I understood the intend correctly)

dags/gmail_data_import_pipeline.py
```diff
@@ -148,7 +150,63 @@ def gmail_link_message_thread_ids_etl(**kwargs):
...
+                                [df_thread_details, df_thread_details_temp],
+                                ignore_index=True
+                            )
+
+    with TemporaryDirectory() as tmp_dir:
```
de-code, line 187: It almost seems like this block appears three times. Maybe it would be worth refactoring.

Just mentioning it in case it's useful. In `data-science-dags` there is a [to_gbq method](https://github.com/elifesciences/data-science-dags/blob/46559fa41d8641137981d1490466e4e7363d0a3a/data_science_pipeline/utils/bq.py#L200) that is more or less the same as pandas' [to_gbq](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_gbq.html), except it uses our schema handling (well a copy of it since it's in another repo).

data_pipeline/gmail_data/get_gmail_data.py
HazalCiplak: Thank you for mentioning about this usage of `concat`!

data_pipeline/gmail_data/get_gmail_data.py
HazalCiplak: I am not sure about that, maybe tomorrow I can show you the data and we can decide together. 
Because the data here is complex and duplicated many times, if we get as dump the data unnecessarily will be in our DB. If there is not a huge difference with the performance, preferring getting summary information would be easier to manage.
And because of having a fixed schema it will be difficult to manage if get this nested json dump.

de-code: Okay, good idea. Let's discuss it later.

Just noting down some benefits I would see with dumping the data as is (i.e. doing more ELT instead of ETL):
- pipeline would be more generic
- it keeps the business logic in one place (bigquery views), although those have admittedly the issue of being more difficult to test
- it will be easier to apply different logic by creating another view, rather than having to retrieve the data again anf filtering it in a different way
- it will provide more data we might be interested later, such as the body text (although it's fair to argue that we can worry about it when it is needed)
- it will make it easier to update as you could just add new messages to a thread and then aggregate in BigQuery (now you would need to retrieve all messages of a thread in order to update the thread summary information)

Regarding the schema, I would be interested to get he auto-schema to work (or at least understand why it wasn't working). Although there may still be an argument to specify it manually.
We could also select the fields as I think you are doing for the labels.

dags/gmail_data_import_pipeline.py
```diff
@@ -139,27 +147,74 @@ def gmail_link_message_thread_ids_etl(**kwargs):
...
+        table_name=data_config.table_name_list_of_thread_ids
+        )
+
+    df_thread_details = pd.concat(
+        [get_one_thread(
+            get_gmail_service(),
+            user_id,
+            id)
+            for id in df_thread_id_list[0]]
+        )
```
de-code: minor: indentation (would move the square bracket here and the round bracket to align)

dags/gmail_data_import_pipeline.py
```diff
@@ -139,27 +147,74 @@ def gmail_link_message_thread_ids_etl(**kwargs):
...
+    table_name = data_config.table_name_thread_details
+    dataset_name = data_config.dataset
+    project_name = data_config.project_name
+
+    df_thread_id_list = get_distinct_values_from_bq(
+        project_name=data_config.project_name,
+        dataset=dataset_name,
+        column_name=data_config.column_name_list_of_thread_ids,
+        table_name=data_config.table_name_list_of_thread_ids
+        )
```
de-code: minor: indentation

data_pipeline/gmail_data/get_gmail_data.py
```diff
@@ -46,14 +52,56 @@ def iter_link_message_thread(service: Resource, user_id: str) -> Iterable[dict]:
...
+@backoff.on_exception(backoff.expo, TimeoutError, max_tries=10)
+def get_one_thread(service: str, user_id: str, thread_id: str) -> pd.DataFrame:
+    imported_timestamp = get_current_timestamp()
+    thread_results = service.users().threads().get(userId=user_id, id=thread_id).execute()
+
+    df_thread = pd.DataFrame()
+
+    df_thread['threadId'] = [thread_results['id']]
+    df_thread['historyId'] = [thread_results['historyId']]
+    df_thread['total_email_count'] = [len(thread_results['messages'])]
```
de-code: minor: just adding a comment anyway, that it seems redundant since we can get the count via the messages (and it would update if add new messages). But you might come across that when looking into getting incremental data (#151)

de-code (approved):

dags/gmail_data_import_pipeline.py
HazalCiplak: Thank you for mentioning that, it can be considered for next version of this pipeline. For now, because of generating schema from the file it is easy to continue using files.

de-code:
Would be good to start adding tests at some point :-)

PR #413, opened 2021-03-25
excluded ids already have details

HazalCiplak:
For the initial load the id list in details excluded when the history details added updated ids will be added in the list to execute

### DIFF
--- a/data_pipeline/gmail_data/get_gmail_data.py
+++ b/data_pipeline/gmail_data/get_gmail_data.py
@@ -85,19 +85,36 @@ def write_dataframe_to_jsonl_file(
         df_data_to_write.to_json(file, orient='records', lines=True, force_ascii=False)
 
 
+# pylint: disable=duplicate-string-formatting-argument
 @backoff.on_exception(backoff.expo, TimeoutError, max_tries=10)
 def get_distinct_values_from_bq(
             project_name: str,
             dataset: str,
             column_name: str,
-            table_name: str
+            table_name_to_execute: str,
+            table_name_executed: str
         ) -> pd.DataFrame:
 
     sql = (
         """
         SELECT DISTINCT {}
         FROM  `{}.{}.{}`
-        """.format(column_name, project_name, dataset, table_name)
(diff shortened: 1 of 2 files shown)


data_pipeline/gmail_data/get_gmail_data.py
```diff
@@ -85,19 +85,36 @@ def write_dataframe_to_jsonl_file(
...
 
 
+# pylint: disable=duplicate-string-formatting-argument
 @backoff.on_exception(backoff.expo, TimeoutError, max_tries=10)
 def get_distinct_values_from_bq(
             project_name: str,
             dataset: str,
             column_name: str,
-            table_name: str
+            table_name_to_execute: str,
```
de-code: I am slightly confused by the new names. Initially I thought it was a type and was meant to be `...to_exclude`. But then this seems to be the source table whereas the next one is used for exclusion?

data_pipeline/gmail_data/get_gmail_data.py
```diff
@@ -85,19 +85,36 @@ def write_dataframe_to_jsonl_file(
...
+            table_name_executed: str
         ) -> pd.DataFrame:
 
     sql = (
         """
         SELECT DISTINCT {}
         FROM  `{}.{}.{}`
-        """.format(column_name, project_name, dataset, table_name)
+        WHERE {} NOT IN
```
de-code: with that many parameters it might be worth using named parameters (it's a bit difficult to follow where what parameter is inserted). Then you also wouldn't need to pass in the same parameter twice I think.

de-code: this could also be a new opportunity for a unit test, testing the generation of the correct SQL :-)

data_pipeline/gmail_data/get_gmail_data.py
HazalCiplak: I am changing the names

de-code (approved):
I would suggest to review the names and use named parameters. Otherwise looks good.

PR #420, opened 2021-03-29
Gmail incremental load

HazalCiplak:
https://github.com/elifesciences/data-hub-issues/issues/151

### DIFF
--- a/data_pipeline/gmail_data/get_gmail_data.py
+++ b/data_pipeline/gmail_data/get_gmail_data.py
@@ -117,7 +117,7 @@ def get_distinct_values_from_bq(
             project_name: str,
             dataset_name: str,
             column_name: str,
-            table_name: str,
+            table_name_source: str,
             table_name_for_exclusion: str,
             table_name_for_update: str
         ) -> pd.DataFrame:
@@ -125,7 +125,7 @@ def get_distinct_values_from_bq(
     sql = (
         """
         SELECT DISTINCT {column_name} AS column
-        FROM  `{project_name}.{dataset_name}.{table_name}`
+        FROM  `{project_name}.{dataset_name}.{table_name_source}`
         WHERE {column_name} NOT IN
             (
                 SELECT {column_name}
@@ -139,7 +139,7 @@ def get_distinct_values_from_bq(
                 column_name=column_name,
                 project_name=project_name,
                 dataset_name=dataset_name,
-                table_name=table_name,
+                table_name_source=table_name_source,
                 table_name_for_exclusion=table_name_for_exclusion,
                 table_name_for_update=table_name_for_update
             )
--- a/data_pipeline/gmail_data/get_gmail_data.py
+++ b/data_pipeline/gmail_data/get_gmail_data.py
@@ -94,11 +94,10 @@ def get_gmail_history_details(service: Resource, user_id: str,  start_id: str) -
     imported_timestamp = get_current_timestamp()
     df_temp = pd.DataFrame(iter_gmail_history(service, user_id, start_id))
     df_hist = pd.DataFrame()
+    df_temp = df_temp.explode('messages')
     df_hist['historyId'] = df_temp['id']
-    thread_list = []
-    for message in df_temp['messages']:
-        thread_list.append(message['threadId'])
-    df_hist['threadId'] = [thread_list]
+    df_hist['messages'] = df_temp['messages']
+    df_hist['threadId'] = df_hist['messages'].apply(pd.Series)['threadId']
     df_hist['user_id'] = user_id
     df_hist['imported_timestamp'] = imported_timestamp
     return df_hist
--- a/data_pipeline/gmail_data/get_gmail_data.py
+++ b/data_pipeline/gmail_data/get_gmail_data.py
@@ -96,7 +96,10 @@ def get_gmail_history_details(service: Resource, user_id: str,  start_id: str) -
     df_temp = pd.DataFrame(iter_gmail_history(service, user_id, start_id))
     df_hist = pd.DataFrame()
     df_hist['historyId'] = df_temp['id']
-    df_hist['messages'] = df_temp['messages']
+    thread_list=[]
+    for message in df_temp['messages']:
+        thread_list.append(message['threadId'])
+    df_hist['threadId'] = [thread_list]
     df_hist['user_id'] = user_id
     df_hist['imported_timestamp'] = imported_timestamp
     return df_hist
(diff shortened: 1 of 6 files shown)


HazalCiplak:
Current task dependencies:

Screenshot 2021-03-30 at 13 49 47

de-code:
Thank you for explaining it patiently.

HazalCiplak:
Task dependencies have changed as the screenshot below. Hope it is more clear.

Screenshot 2021-03-30 at 14 54 54

HazalCiplak:
Please let me know if you have any question about the implementation. We can look at it together.

de-code:
Thank you, that already looks easier to understand.

One thing I still don't fully understand is...
What is the purpose of `load_new_data_from_temp_table_thread_ids`? As it is not obvious what this is feeding into.

Then there is still the question regarding the initial schema creation. Which is also related to still some uncertainty around `gmail_history_details_to_temp_table_etl`.
Just an idea, what if the pipeline was like:
`gmail_thread_ids_list_to_temp_table_etl` -> `gmail_thread_details_from_temp_thread_ids_etl` -> `gmail_history_details_to_temp_table_etl` -> `gmail_thread_details_from_temp_history_ids_etl` ...
Would that make any sense?

Deleting tables is fine for now, it may be something to review when (or if) we add another email address to ingest.

Perhaps we could go through it again. I would just like to make sure that I understand the dependencies correctly. (Apologies for being slow to fully understand it)

HazalCiplak:
> What is the purpose of `load_new_data_from_temp_table_thread_ids`? As it is not obvious what this is feeding into.

Have changed the name of function `load_from_temp_table_to_thread_ids_list`, It is loading new comming thread ids data from temporary table to actual table which's name '`gmail_thread_ids_list`'. Does this change make it more clear?

> Then there is still the question regarding the initial schema creation. Which is also related to still some uncertainty around `gmail_history_details_to_temp_table_etl`.
> Just an idea, what if the pipeline was like:
> `gmail_thread_ids_list_to_temp_table_etl` -> `gmail_thread_details_from_temp_thread_ids_etl` -> `gmail_history_details_to_temp_table_etl` -> `gmail_thread_details_from_temp_history_ids_etl` ...
> Would that make any sense?

Yes it makes sense! Thank you for the idea. Let me change this part like that.

> Deleting tables is fine for now, it may be something to review when (or if) we add another email address to ingest.
> Perhaps we could go through it again. I would just like to make sure that I understand the dependencies correctly. (Apologies for being slow to fully understand it)

No worries about being slow, It helps improving the structure 😊

de-code:
> > What is the purpose of `load_new_data_from_temp_table_thread_ids`? As it is not obvious what this is feeding into.
> 
> Have changed the name of function `load_from_temp_table_to_thread_ids_list`, It is loading new comming thread ids data from temporary table to actual table which's name '`gmail_thread_ids_list`'. Does this change make it more clear?

I understand that it is populating the de-duplicated table of thread ids. What do we do with the contents of that table? Is it need for Data Studio or another another task? (I couldn't see `data_config.table_name_thread_ids` being used in another task)

HazalCiplak:
> > > What is the purpose of `load_new_data_from_temp_table_thread_ids`? As it is not obvious what this is feeding into.
> > 
> > 
> > Have changed the name of function `load_from_temp_table_to_thread_ids_list`, It is loading new comming thread ids data from temporary table to actual table which's name '`gmail_thread_ids_list`'. Does this change make it more clear?
> 
> I understand that it is populating the de-duplicated table of thread ids. What do we do with the contents of that table? Is it need for Data Studio or another another task? (I couldn't see `data_config.table_name_thread_ids` being used in another task)

Ah I see. This table basically keeping threadId and messageId relationships, it will be needed in case of creating the modelled DWH.

de-code:
> Ah I see. This table basically keeping threadId and messageId relationships, it will be needed in case of creating the modelled DWH.

I see. Do  we maybe already have the information from the thread details? But it might become more clear later.

HazalCiplak:
> > Ah I see. This table basically keeping threadId and messageId relationships, it will be needed in case of creating the modelled DWH.
> 
> I see. Do we maybe already have the information from the thread details? But it might become more clear later.

Yes we already have this information but it is sort of data lake structure, so keeping that table too will be more suitable for our DWH model

data_pipeline/gmail_data/get_gmail_data.py
```diff
@@ -96,7 +95,10 @@ def get_gmail_history_details(service: Resource, user_id: str,  start_id: str) -
     df_temp = pd.DataFrame(iter_gmail_history(service, user_id, start_id))
     df_hist = pd.DataFrame()
     df_hist['historyId'] = df_temp['id']
-    df_hist['messages'] = df_temp['messages']
+    thread_list = []
+    for message in df_temp['messages']:
+        thread_list.append(message['threadId'])
```
de-code: minor: you probably know you could use a list comprehension here

data_pipeline/gmail_data/get_gmail_data.py
```diff
@@ -96,7 +95,10 @@ def get_gmail_history_details(service: Resource, user_id: str,  start_id: str) -
...
+    df_hist['threadId'] = [thread_list]
```
de-code: if each value is a list (array in BigQuery), then maybe name it `threadIds`?

data_pipeline/utils/data_store/bq_data_service.py
```diff
@@ -349,3 +351,70 @@ def load_new_data_in_temp_table_to_actual_table(
...
+            dataset_name: str,
+            column_name: str,
+            table_name_source: str,
+            table_name_for_exclusion: str = None
+        ) -> pd.DataFrame:
+
+    sql = """
+        SELECT DISTINCT {{ column_name }}
+        FROM  `{{ project_name }}.{{ dataset_name }}.{{ table_name_source }}`
+        {% if table_name_for_exclusion %}
```
de-code, line 367: comment: initially I thought you gone fancy with JinjaSql, but this seems a quite nice use-case for it

data_pipeline/utils/data_store/bq_data_service.py
```diff
@@ -349,3 +351,70 @@ def load_new_data_in_temp_table_to_actual_table(
...
+        'table_name_for_exclusion': table_name_for_exclusion
+        }
+
+    jin = JinjaSql(param_style='pyformat')
+    query, bind_params = jin.prepare_query(sql, params)
+    client = get_bq_client(project=project_name)
+    query_job = client.query(query % bind_params)  # Make an API request.
+    results = query_job.result()  # Waits for query to finish
+
+    return pd.concat([pd.DataFrame([row.column]) for row in results], ignore_index=True)
```
de-code: minor: now that this is "shared", would it now be worth to just very slightly improve this line?
personally I would just return a list of values, i.e.

```python
return [row.column for row in results]
```

If you want to use Pandas, then Series seems to be the more natural choice as it's just one list of values, e.g.:

```python
return pd.Series([row.column for row in results])
```

Or with a DataFrame (but it wouldn't be obvious what the column name in the result is):

```python
return pd.Series([row.column for row in results]).to_frame(column_name)
```

data_pipeline/utils/data_store/bq_data_service.py
```diff
@@ -349,3 +351,70 @@ def load_new_data_in_temp_table_to_actual_table(
...
+def get_max_value_from_bq_table(
+            project_name: str,
+            dataset_name: str,
+            column_name: str,
+            table_name: str,
+        ) -> str:
+
+    sql = """
+        SELECT
+        MAX({{ column_name }}) AS start_id
```
de-code: minor: `start_id` could maybe more generic now, e.g. `value`?

data_pipeline/utils/data_store/bq_data_service.py
```diff
@@ -349,3 +351,70 @@ def load_new_data_in_temp_table_to_actual_table(
...
+        'column_name': column_name
+        }
+
+    jin = JinjaSql(param_style='pyformat')
+    query, bind_params = jin.prepare_query(sql, params)
+    client = get_bq_client(project=project_name)
+    query_job = client.query(query % bind_params)  # Make an API request.
+    results = query_job.result()  # Waits for query to finish
+
+    for row in results:
```
de-code: minor: now that this is "shared", I would also suggest to just improve it very slightly. e.g. in this case the result will be exactly one row. So looping to me makes it a bit more difficult to read. I would also not convert it to a string. If that was desired, the caller could easily do it.

e.g.:

```python
return list(results)[0].value
```

de-code (approved):
just some minor comments. Looking good.

HazalCiplak:
The final dependencies:

Screenshot 2021-03-30 at 21 10 29

PR #462, opened 2021-04-27
Gmail end2end test

HazalCiplak:
https://github.com/elifesciences/data-hub-issues/issues/165

### DIFF
--- a/dags/gmail_data_import_pipeline.py
+++ b/dags/gmail_data_import_pipeline.py
@@ -49,6 +49,8 @@ GMAIL_DATA_CONFIG_FILE_PATH_ENV_NAME = "GMAIL_DATA_CONFIG_FILE_PATH"
 DEPLOYMENT_ENV_ENV_NAME = "DEPLOYMENT_ENV"
 DEFAULT_DEPLOYMENT_ENV = "ci"
 
+IS_GMAIL_END2END_TEST_ENV = "IS_GMAIL_END2END_TEST"
+
 DAG_ID = "Gmail_Data_Import_Pipeline"
 
 CHUNK_SIZE = 3000
@@ -89,6 +91,10 @@ def data_config_from_xcom(context):
 
 def get_gmail_user_id():
     return get_env_var_or_use_default(GMAIL_DATA_USER_ID_ENV)
+
+
+def is_gmail_end2end_test():
+    return get_env_var_or_use_default(IS_GMAIL_END2END_TEST_ENV)
 
 
 def get_gmail_service():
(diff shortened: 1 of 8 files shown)


dags/gmail_data_import_pipeline.py
```diff
@@ -91,6 +93,10 @@ def get_gmail_user_id():
     return get_env_var_or_use_default(GMAIL_DATA_USER_ID_ENV)
 
 
+def is_gmail_end2end_test():
+    return get_env_var_or_use_default(IS_GMAIL_END2END_TEST_ENV)
```
de-code: This could potentially cause confusion. e.g. if you set the environment variable to `False`, then it will still think that we are running end-to-end tests. That is because it is just treating it as a string.

Python seem to be missing a simple `to_bool` function. So you could use something like: `value in {'True'}` (not sure if there is already a helper function in the project).

HazalCiplak: I found a way, maybe it can work:
`bool(distutils.util.strtobool(some_string))`
https://docs.python.org/3/distutils/apiref.html#distutils.util.strtobool

HazalCiplak: Or we can create our own function like:

```
def str2bool(value):
  return value.lower() in ("yes", "y", "true", "t", "1")
```

de-code: I think I have used `strtobool` from `distutils` before.
It is a bit of a shame that it is not part of the more core Python API. i.e. it feels like we shouldn't use it, but it does what we want to do, except that it should return a boolean rather than `1` or `0` (but you might prefer that ;-) ).
The other thing that it does perhaps differently but could be handled is what to do with `None` or blank strings.

In any case, we are in luck. This project already has `str_to_bool`: https://github.com/elifesciences/data-hub-core-airflow-dags/blob/c4b0cda4a9096e167a370dce12742519c5f960a6/data_pipeline/utils/pipeline_config.py#L70

Although still no option of providing a default value directly. Perhaps it could be extended.

data_pipeline/utils/pipeline_config.py
```diff
@@ -64,7 +64,7 @@ def replace_env_placeholder(
 
 
 TRUE_VALUES = {'true', '1', 't', 'y', 'yes'}
-FALSE_VALUES = {'false', '0', 'f', 'n', 'no'}
```
de-code, line 67: Personally I think it might still be better to explicitly require a default value to be passed in and raise an issue otherwise.

e.g. something like:

```python
def str_to_bool(value: str, default_value: Optional[bool] = None) -> bool:
    if not value and default_value is not None:
        return default_value
    lower_value = value.lower()
    if lower_value in TRUE_VALUES:
        return True
    if lower_value in FALSE_VALUES:
        return False
    raise ValueError('unrecognised boolean value: %r' % value)
```

It also makes it easier to see what the behaviour should be. e.g.

```python
# No default value, makes it more unclear what a user would expect. By default I would maybe expect an error.
str_to_bool('')

# seeing the default value makes it clear what the intended behaviour is
str_to_bool('', default_value=False)
```

de-code: Maybe also worth noting, it appears that `get_env_var_or_use_default` seems to have a default value of `None` (unless specified). So the above might still break if the environment variable wasn't defined?

de-code (approved):

PR #479, opened 2021-05-06
Remove duplication of gmail code

HazalCiplak:
https://github.com/elifesciences/data-hub-issues/issues/161

### DIFF
--- a/dags/gmail_data_import_pipeline.py
+++ b/dags/gmail_data_import_pipeline.py
@@ -122,7 +122,7 @@ def get_gmail_df_to_load_bq_table(
         df_data_to_write: pd.DataFrame):
 
     with TemporaryDirectory() as tmp_dir:
-        filename = os.path.join(tmp_dir, table_name+'.json')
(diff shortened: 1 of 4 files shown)


dags/gmail_data_import_pipeline.py
```diff
@@ -116,129 +115,94 @@ def get_gmail_service():
     )
 
 
-def gmail_label_data_to_temp_table_etl(**kwargs):
-    data_config = data_config_from_xcom(kwargs)
-    user_id = get_gmail_user_id()
-    project_name = data_config.project_name
-    dataset_name = data_config.dataset_name
-    table_name = data_config.temp_table_name_labels
+def get_gmail_df_to_load_bq_table(
```
de-code: Maybe we could try to find a different name. Just because `get` usually implies it is returning something and also wouldn't usually have a side effect... maybe turning it around to something like: `load_bq_table_from_df`?

de-code (approved):

PR #495, opened 2021-05-17
added retry to the etl task

HazalCiplak:
https://github.com/elifesciences/data-hub-issues/issues/196

### DIFF
--- a/dags/crossref_event_data_import_pipeline.py
+++ b/dags/crossref_event_data_import_pipeline.py
@@ -207,7 +207,7 @@ create_table_if_not_exist_task = create_python_task(
 )
 crossref_event_data_etl_task = create_python_task(
     CROSSREF_DAG, "crossref_event_data_etl",
-    crossref_data_etl, retries=2
+    crossref_data_etl, retries=5
 )
 log_last_record_date_task = create_python_task(
     CROSSREF_DAG,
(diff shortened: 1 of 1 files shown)


dags/crossref_event_data_import_pipeline.py
```diff
@@ -207,7 +207,7 @@ def log_last_record_date(**kwargs):
 )
 crossref_event_data_etl_task = create_python_task(
     CROSSREF_DAG, "crossref_event_data_etl",
-    crossref_data_etl,
```
de-code, line 210: might be slightly easier to read as a separate line (although it is the same above).

Just making a note that one can also set `default_args` on a DAG, that are applied to the tasks.
I am not sure whether it is an advantage or not. It might help keeping the DAG focused.

I used it in Data Science DAGs, as part of `create_dag`: https://github.com/elifesciences/data-science-dags/blob/455eb44b66f04b673bb97475dfa731d1e96a7963/dags/peerscout_recommend_senior_editors.py#L29-L31

Defined in: https://github.com/elifesciences/data-science-dags/blob/455eb44b66f04b673bb97475dfa731d1e96a7963/data_science_pipeline/utils/dags.py#L53-L89

de-code (approved):

dags/crossref_event_data_import_pipeline.py
HazalCiplak: OK, I am separating them in lines.
Thank you for sharing `default_arg`, I will be helpful in case of a lot of tasks in a DAG I will use it then.

PR #506, opened 2021-05-24
Remove fix data format from generic web api

HazalCiplak:
https://github.com/elifesciences/data-hub-issues/issues/172

### DIFF
--- a/data_pipeline/utils/data_pipeline_timestamp.py
+++ b/data_pipeline/utils/data_pipeline_timestamp.py
@@ -2,8 +2,8 @@ from datetime import (
     datetime,
     timezone
 )
+import dateparser
 import pytz
-from dateutil import parser
 
 
 def get_current_timestamp_as_string(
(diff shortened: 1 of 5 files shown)


data_pipeline/utils/data_pipeline_timestamp.py
```diff
@@ -26,8 +26,8 @@ def parse_timestamp_from_str(timestamp_as_str, time_format: str = None):
             timestamp_as_str.strip(), time_format
         )
     else:
-        timestamp_obj = dateparser.parse(
-            timestamp_as_str.strip()
+        timestamp_obj = parser.parse(
```
de-code: just some minor comments:

`dateutil` doesn't seem to be in the `requirements.txt` file and we should probably add it if we are using. (and remove `dateparser` if we don't use it anymore)
I would maybe suggest to refer to it here as `dateutil.parser` simply because `parser`is quite generic and it might help someone reading the code, to understand what we are using (without scrolling).

The advantage and disadvantage of `dateutil` is that it is very forgiving. i.e. it will attempt to parse anything you throw at it, but may not alert us if there is something wrong. It is certainly convenient though.

de-code (approved):

PR #593, opened 2021-07-21
Add data hub monitoring

HazalCiplak:
https://github.com/elifesciences/data-hub-issues/issues/239
https://github.com/elifesciences/data-hub-issues/issues/238


de-code:
I had thought we would put a `monitoring` package underneath `data_pipeline`.
For PeerScout that was different, because it is deployed separately and has its own dependencies.
What advantages do you see in having it on its own at the top level?
(moving it inside `data_pipeline` would probably simplify a few things)

monitoring/data_hub_pipeline_health_check.py
```diff
@@ -0,0 +1,132 @@
...
+        table=table
+    )
+
+
+def read_dataframe_from_s3_bucket(
+    bucket_name: str,
+    object_name: str
+) -> pd.DataFrame:
+    s3_object = boto3.client('s3').get_object(Bucket=bucket_name, Key=object_name)
+    return pd.read_csv(BytesIO(s3_object['Body'].read()))
```
de-code: whenever there is a `read` from a stream, I am thinking it probably needs a context manager to ensure the stream is closed afterwards.

There are BTW some S3 related utilites in https://github.com/elifesciences/data-hub-core-airflow-dags/blob/develop/data_pipeline/utils/data_store/s3_data_service.py
Could also add a `read_bytes_from_s3` or similar.
Pandas can also read from s3 if you give it a URL, e.g. `s3://bucket/object` (it will need the related `fsspec` dependencies)

monitoring/ping_healthchecks.py
```diff
@@ -0,0 +1,25 @@
+import os
+import logging
+import socket
+import requests
+
+LOGGER = logging.getLogger(__name__)
+
+HEALTH_CHECK_URL_ENV = "HEALTH_CHECK_URL"
+DEFAULT_HEALTH_CHECK_URL_ENV = "https://hc-ping.com/3549bc53-4bd9-4ef4-9c4a-22aa57c2fb5b"
```
de-code: minor name: probably should be `DEFAULT_HEALTH_CHECK_URL` (without the `_ENV`)

Is this a development healtheck url and we will use a different one in production?

monitoring/ping_healthchecks.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def ping():
+    url = get_env_var_or_use_default(
+        HEALTH_CHECK_URL_ENV, DEFAULT_HEALTH_CHECK_URL_ENV)
+    LOGGER.info('[healthcheck] pinging url: %s', url)
+    try:
+        requests.post(url)
+    except socket.error as err:
+        LOGGER.info("Ping failed: %s", err)
```
de-code: Should we maybe re-raise the exception?
Otherwise it is saying "successful" afterwards.

A post may also get sent to the server but we are getting a server error. calling [raise_for_status](https://docs.python-requests.org/en/master/api/#requests.Response.raise_for_status) is usually good.

requirements.monitoring.txt
```diff
@@ -0,0 +1,7 @@
+numpy==1.18.2
+pandas==1.0.3
+pandas-gbq==0.13.1
+pyarrow==3.0.0
+tqdm==4.45.0
+slackclient==2.5.0
```
de-code: apologies, the list should have been pruned. I don't think we are using `slackclient`.

dags/monitor_data_hub_pipeline_health.py
```diff
@@ -0,0 +1,127 @@
...
+    LOGGER.info('data_config_dict: %s', data_config_dict)
+    kwargs["ti"].xcom_push(
+        key="data_config_dict",
+        value=data_config_dict
+    )
+
+
+# pylint: disable=unused-argument
+def ping_health_checks_io(**context):
+    logging.basicConfig(level='INFO')
```
de-code: The logging initialisation should really only happen in an application's main function. Within Airflow, logging should already be configured.

dags/monitor_data_hub_pipeline_health.py
```diff
@@ -0,0 +1,127 @@
...
+    LOGGER.info('conf_file_path: %s', conf_file_path)
+    data_config_dict = get_yaml_file_as_dict(conf_file_path)
+    LOGGER.info('data_config_dict: %s', data_config_dict)
+    kwargs["ti"].xcom_push(
+        key="data_config_dict",
+        value=data_config_dict
+    )
+
+
+# pylint: disable=unused-argument
```
de-code: just a comment: you can rename the parameter to `_` or `__`. That seems to be a common approach to avoid the linting error.

dags/monitor_data_hub_pipeline_health.py
```diff
@@ -0,0 +1,127 @@
...
+MONITOR_DATA_HUB_PIPELINE_HEALTH_SCHEDULE_INTERVAL_ENV_NAME = (
+    "MONITOR_DATA_HUB_PIPELINE_HEALTH_SCHEDULE_INTERVAL"
+)
+
+MONITOR_DATA_HUB_PIPELINE_HEALTH_DAG = create_dag(
+    dag_id=DAG_ID,
+    # schedule_interval=os.getenv(
+    #     MONITOR_DATA_HUB_PIPELINE_HEALTH_SCHEDULE_INTERVAL_ENV_NAME
+    # ),
+    dagrun_timeout=timedelta(hours=6)
```
de-code: just leaving a comment that this needs to be changed before merging (you were probably planning to anyway).

dags/monitor_data_hub_pipeline_health.py
```diff
@@ -0,0 +1,127 @@
...
+)
+
+
+def get_env_var_or_use_default(env_var_name, default_value=None):
```
de-code: might be worth moving to a common module? (the same function is defined for the gmail dag, there are also other uses of `os.getenv` but I wouldn't suggest to change those.)

dags/monitor_data_hub_pipeline_health.py
```diff
@@ -0,0 +1,127 @@
...
+)
+
+check_data_hub_tables_status_task = create_python_task(
+    MONITOR_DATA_HUB_PIPELINE_HEALTH_DAG,
+    "check_data_hub_tables_status",
+    check_data_hub_tables_status,
+    retries=5
+)
+
+# pylint: disable=pointless-statement
```
de-code: also just a comment: A common pattern I have seen for that is to assign it to underscore, e.g.: `_ = ( .... )`

Dockerfile
```diff
@@ -14,6 +14,9 @@ USER airflow
 COPY requirements.build.txt ./
 RUN pip install --disable-pip-version-check -r requirements.build.txt
 
+COPY requirements.monitoring.txt ./
+RUN pip install --disable-pip-version-check -r requirements.monitoring.txt
```
de-code, line 18: Maybe just add any requirements you may need for monitoring to `requirements.txt` or do you see an advantage in keeping it separate?

The `requirements.build.txt` for example are separate because the version of those dependencies (including `pip`) can affect the installation of the other dependencies.

dags/monitor_data_hub_pipeline_health.py
```diff
@@ -0,0 +1,122 @@
...
+    dag_context = context["ti"]
+    data_config_dict = dag_context.xcom_pull(
+        key="data_config_dict", task_ids="get_data_config"
+    )
+    LOGGER.info('data_config_dict: %s', data_config_dict)
+    deployment_env = get_env_var_or_use_default(
+        DEPLOYMENT_ENV_ENV_NAME, DEFAULT_DEPLOYMENT_ENV)
+    data_config = MonitoringConfig(
+        data_config_dict, deployment_env)
+    LOGGER.info('data_config: %r', data_config)
```
de-code, line 58: just mentioning that this won't log anything useful apart from that `data_config` is not None (or what type it is). That is because `MonitoringConfig` doesn't implement `__str__` or `__repr__`

data_pipeline/monitoring/monitoring_config.py
```diff
@@ -0,0 +1,22 @@
+from data_pipeline.utils.pipeline_config import (
+    update_deployment_env_placeholder
+)
+
+
+# pylint: disable=too-few-public-methods, too-many-instance-attributes
+class MonitoringConfig:
+
+    def __init__(self, data_config: dict,
```
de-code, line 9: This is how we have done it for other config. So it is great for consistency.
I am just putting this out there as a potential option for the future...
Perhaps consider making it a `NamedTuple`. Those seem to look neater and they for example come with a `__repr__` implementation.

Something like:

```python
class MonitoringConfig(NamedTuple):
    project_name: str
    dataset_name: str
    table_name: str
    bucket_name: str
    object_name: str

    @staticmethod
    def from_dict(data_config: dict, ...):
        resolved_data_config = update_deployment_env_placeholder(...)
        return MonitoringConfig(
            project_name=resolved_data_config.get('project_name'),
            dataset_name=resolved_data_config.get('dataset_name'),
            ...
        )
```

The other thing that might be worth considering is, what configuration options are required? I think we came across that during the gmail development, when a value was misconfigured and caused a None exception later on, rather than when loading the config. Using `data_config['project']` would at least require it to be set (not necessarily being non-empty).

Dockerfile
HazalCiplak: Because the requirements list is very long already I wanted to keep it separate in case of we want to move monitoring in another repo in the future as it is not a data pipeline. 
It was the same idea for me to keep monitoring as a separate folder at the beginning.

dags/monitor_data_hub_pipeline_health.py
HazalCiplak: Himm I can add this functions but when you come back we need to talk about it :D as it is not clear for me.

HazalCiplak:
I am merging the code, if there is something missing Daniel we can check the in next PRs.

de-code:
> I am merging the code, if there is something missing Daniel we can check the in next PRs.

Apologies, I was meant to approve it before I was off.

PR #615, opened 2021-08-02
Remove default slack url

HazalCiplak:
https://github.com/elifesciences/data-hub-issues/issues/240


data_pipeline/monitoring/data_hub_pipeline_health_check.py
```diff
@@ -140,7 +135,8 @@ def run_data_hub_pipeline_health_check(
             changed_status_df = get_changed_status_df(merged_status_df)
             if len(changed_status_df) > 0:  # pylint: disable=len-as-condition
                 LOGGER.info('changed: %s', changed_status_df.to_dict(orient='rows'))
-                send_slack_notification(changed_status_df)
+                if deployment_env in ("staging", "prod"):
```
de-code: I would recommend making this conditional on whether a url is configured. Then it can still be tested in dev by providing such url. It will also make it easier to for example setup the monitoring in another environment, such as the flux cluster. (It is also better not to have deployment specific values in t he code)

de-code (approved):

PR #769, opened 2021-11-16
Get surveymonkey survey list

HazalCiplak:
https://github.com/elifesciences/data-hub-issues/issues/281


docker-compose.yml
```diff
@@ -38,12 +39,14 @@ x-airflow-env:
     - MATERIALIZE_BIGQUERY_VIEWS_VIEW_MAPPING_ENABLED=${MATERIALIZE_BIGQUERY_VIEWS_VIEW_MAPPING_ENABLED}
     - MONITORING_CONFIG_FILE_PATH=/home/airflow/app-config/monitoring/monitoring.config.yaml
     - HEALTH_CHECK_URL=https://hc-ping.com/3549bc53-4bd9-4ef4-9c4a-22aa57c2fb5b
+    - SURVEYMONKEY_ACCESS_TOKEN='SURVEYMONKEY_ACCESS_TOKEN'
```
de-code: Secret files seem to be slightly more secure than secret environment variables.

HazalCiplak: Yes, sure. I have add it for now It is not working in this way actually :D I will add sealed secret in flux after deploy.

data_pipeline/surveymonkey/surveymonkey_etl.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def get_connection_surveymonkey_api() -> http:
+    return http.client.HTTPSConnection("api.surveymonkey.com")
+
+
+def get_surveymonkey_api_headers() -> dict:
+    return {
+        'Accept': "application/json",
+        'Authorization': f"Bearer {SURVEYMONKEY_ACCESS_TOKEN_ENV_VAR}"
```
de-code: I guess here you will need to get the actual token, not the env var (or filename to the token).
Or rather I mean you need the token at this point. But probably get it outside this function.

HazalCiplak: Ah yes I missed! Thank you

data_pipeline/surveymonkey/surveymonkey_etl.py
```diff
@@ -0,0 +1,30 @@
...
+        }
+
+
+def get_survey_list() -> list:
+    conn = get_connection_surveymonkey_api()
+    headers = get_surveymonkey_api_headers()
+
+    conn.request("GET", "/v3/surveys", headers=headers)
```
de-code: Just a thought:

Using the `http` library seems quite manual. I am not even sure whether that is actually using HTTPS?
I find the `requests` library much easier to use. e.g.:

```python
import requests

response = requests.get('https://api.surveymonkey.com/v3/surveys', headers=headers)
response.raise_for_status()
return response.json()['data']
```

dags/surveymonkey_data_import_pipeline.py
```diff
@@ -0,0 +1,138 @@
...
+    create_or_extend_table_schema
+)
+
+from data_pipeline.surveymonkey.get_surveymonkey_data_config import (
+    SurveyMonkeyDataConfig
+)
+
+LOGGER = logging.getLogger(__name__)
+
+SURVEYMONKEY_SECRET_FILE_ENV_VAR = "SURVEYMONKEY_SECRET_FILE"
```
de-code: minor: below you call it `ENV_NAME` (I guess it is an env var name, so both are correct)

dags/surveymonkey_data_import_pipeline.py
```diff
@@ -0,0 +1,138 @@
...
+        data_config_dict, deployment_env)
+    LOGGER.info('data_config: %r', data_conf)
+    return data_conf
+
+
+def get_surveymonkey_access_token():
+    secret_file = get_env_var_or_use_default(SURVEYMONKEY_SECRET_FILE_ENV_VAR, "")
+    LOGGER.info("surveymonkey secret file name %s", secret_file)
+    with open(secret_file) as file:
+        return json.load(file)["access_token"]
```
de-code, line 81: just a note: the secret file doesn't need to be json. It could also just be the secret as is.

data_pipeline/surveymonkey/surveymonkey_etl.py
```diff
@@ -0,0 +1,18 @@
...
+import requests
+
+LOGGER = logging.getLogger(__name__)
+
+
+def get_surveymonkey_api_headers(access_token: str) -> dict:
+    return {
+        'Accept': "application/json",
+        'Authorization': f"Bearer {access_token}"
+        }
```
de-code: minor: I'd probably indent that with `return` as that is where it's opened

sample_data_config/surveymonkey/surveymonkey-data-pipeline.config.yaml
```diff
@@ -0,0 +1,4 @@
+projectName: 'elife-data-pipeline'
+datasetName: '{ENV}'
+surveyMonkeySurveyList:
+  table: 'surveymonkey_survey_list'
```
de-code: minor: add line feed at the end?

de-code (approved):

PR #775, opened 2021-11-19
add surveymonkey json transformation for question details

HazalCiplak:
https://github.com/elifesciences/data-hub-issues/issues/283

### DIFF
--- a/sample_data_config/surveymonkey/surveymonkey-data-pipeline.config.yaml
+++ b/sample_data_config/surveymonkey/surveymonkey-data-pipeline.config.yaml
@@ -2,3 +2,6 @@ projectName: 'elife-data-pipeline'
 datasetName: '{ENV}'
 surveyMonkeySurveyList:
   table: 'surveymonkey_survey_list'
+  column: 'id'
+surveyMonkeySurveyQuestions:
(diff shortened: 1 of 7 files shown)


sample_data_config/surveymonkey/surveymonkey-data-pipeline.config.yaml
```diff
@@ -2,3 +2,6 @@ projectName: 'elife-data-pipeline'
 datasetName: '{ENV}'
 surveyMonkeySurveyList:
   table: 'surveymonkey_survey_list'
+  column: 'id'
```
de-code: what is the meaning of that column?

HazalCiplak: ah, that is for getting the `survey id` from bq to get the details for this specific survey. `id` in `survey_list `table.

de-code: okay, maybe it would benefit of a more specific name. e.g. `primary_key_column`?
(it may be worth renaming to `survey_id` for consistency, at some point)

de-code (approved):

PR #778, opened 2021-11-23
Surveymonkey survey answers etl

HazalCiplak:
https://github.com/elifesciences/data-hub-issues/issues/282

### DIFF
--- a/dags/surveymonkey_data_import_pipeline.py
+++ b/dags/surveymonkey_data_import_pipeline.py
@@ -134,8 +134,6 @@ def surveymonkey_survey_answers_etl(**kwargs):
             access_token=get_surveymonkey_access_token(),
             survey_id=survey_id
         )
-        if not iterable_of_answers_of_one_survey:
-            LOGGER.info("There is no answer for the survey %s", survey_id)
 
         load_given_json_list_data_from_tempdir_to_bq(
             project_name=data_config.project_name,
--- a/dags/surveymonkey_data_import_pipeline.py
+++ b/dags/surveymonkey_data_import_pipeline.py
@@ -6,9 +6,11 @@ import json
 from datetime import timedelta
 
 from data_pipeline.surveymonkey.surveymonkey_etl import (
+    get_survey_answer_details,
     get_survey_list,
     get_survey_question_details,
-    get_bq_json_for_survey_questions_response_json
+    get_bq_json_for_survey_questions_response_json,
+    get_bq_json_for_survey_answers_response_json,
 )
 
 from data_pipeline.utils.dags.data_pipeline_dag_utils import (
@@ -102,19 +104,52 @@ def surveymonkey_survey_questions_etl(**kwargs):
     ).values.tolist()
 
     for survey_id in survey_id_list:
+        LOGGER.info("questions for survey_id: %s", str(survey_id[1]))
         survey_questions_list = [get_bq_json_for_survey_questions_response_json(
             get_survey_question_details(
                 access_token=get_surveymonkey_access_token(),
                 survey_id=str(survey_id[1])
             )
         )]
-
         load_given_json_list_data_from_tempdir_to_bq(
             project_name=data_config.project_name,
             dataset_name=data_config.dataset_name,
             table_name=data_config.survey_questions_table_name,
             json_list=survey_questions_list
         )
+
+
+def surveymonkey_survey_answers_etl(**kwargs):
+    data_config = data_config_from_xcom(kwargs)
+    survey_id_list_from_df = get_distinct_values_from_bq(
+        project_name=data_config.project_name,
+        dataset_name=data_config.dataset_name,
+        table_name_source=data_config.survey_list_table_name,
+        column_name=data_config.survey_id_column_name
+    ).values.tolist()
+
+    for survey_id_list in survey_id_list_from_df:
+        survey_id = str(survey_id_list[1])
+        LOGGER.info("answers for survey_id: %s", survey_id)
+        list_of_answers_of_one_survey = get_survey_answer_details(
+            access_token=get_surveymonkey_access_token(),
+            survey_id=survey_id
+        )
+        len_of_answer_list = len(list_of_answers_of_one_survey)
+        LOGGER.info("length of answers is %s", len_of_answer_list)
+        if len_of_answer_list>0:
+            for answers_dict in list_of_answers_of_one_survey:
+                survey_answers_list = [
+                    get_bq_json_for_survey_answers_response_json(answers_dict)
+                ]
+                load_given_json_list_data_from_tempdir_to_bq(
+                    project_name=data_config.project_name,
+                    dataset_name=data_config.dataset_name,
+                    table_name=data_config.survey_answers_table_name,
+                    json_list=survey_answers_list
+                )
+        else:
+            LOGGER.info("There is no answer for the survey %s", survey_id)
 
 
 SURVERMONKEY_DAG = create_dag(
@@ -146,5 +181,20 @@ surveymonkey_survey_questions_task = create_python_task(
     retries=3
 )
 
+surveymonkey_survey_answers_task = create_python_task(
+    SURVERMONKEY_DAG,
+    "surveymonkey_survey_answers_etl",
+    surveymonkey_survey_answers_etl,
+    retries=3
+)
+
 # pylint: disable=pointless-statement
-get_data_config_task >> surveymonkey_survey_list_task >> surveymonkey_survey_questions_task
+(
+    get_data_config_task >> surveymonkey_survey_list_task
+)
+(
+    surveymonkey_survey_list_task >> [
+        surveymonkey_survey_questions_task,
+        surveymonkey_survey_answers_task
+    ]
+)
--- a/dags/surveymonkey_data_import_pipeline.py
+++ b/dags/surveymonkey_data_import_pipeline.py
@@ -137,7 +137,7 @@ def surveymonkey_survey_answers_etl(**kwargs):
         )
         len_of_answer_list = len(list_of_answers_of_one_survey)
         LOGGER.info("length of answers is %s", len_of_answer_list)
-        if len_of_answer_list>0:
+        if len_of_answer_list > 0:
             for answers_dict in list_of_answers_of_one_survey:
                 survey_answers_list = [
                     get_bq_json_for_survey_answers_response_json(answers_dict)
--- a/dags/surveymonkey_data_import_pipeline.py
+++ b/dags/surveymonkey_data_import_pipeline.py
@@ -135,21 +135,19 @@ def surveymonkey_survey_answers_etl(**kwargs):
             access_token=get_surveymonkey_access_token(),
             survey_id=survey_id
         )
-        len_of_answer_list = len(list_of_answers_of_one_survey)
-        LOGGER.info("length of answers is %s", len_of_answer_list)
-        if len_of_answer_list > 0:
-            for answers_dict in list_of_answers_of_one_survey:
-                survey_answers_list = [
-                    get_bq_json_for_survey_answers_response_json(answers_dict)
-                ]
-                load_given_json_list_data_from_tempdir_to_bq(
-                    project_name=data_config.project_name,
-                    dataset_name=data_config.dataset_name,
-                    table_name=data_config.survey_answers_table_name,
-                    json_list=survey_answers_list
-                )
-        else:
+        LOGGER.info("length of answers is %s", len(list_of_answers_of_one_survey))
+        if not list_of_answers_of_one_survey:
             LOGGER.info("There is no answer for the survey %s", survey_id)
+        for answers_dict in list_of_answers_of_one_survey:
+            survey_answers_list = [
+                get_bq_json_for_survey_answers_response_json(answers_dict)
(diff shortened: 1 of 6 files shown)


dags/surveymonkey_data_import_pipeline.py
```diff
@@ -117,6 +119,39 @@ def surveymonkey_survey_questions_etl(**kwargs):
...
+    for survey_id_list in survey_id_list_from_df:
+        survey_id = str(survey_id_list[1])
+        LOGGER.info("answers for survey_id: %s", survey_id)
+        list_of_answers_of_one_survey = get_survey_answer_details(
+            access_token=get_surveymonkey_access_token(),
+            survey_id=survey_id
+        )
+        len_of_answer_list = len(list_of_answers_of_one_survey)
+        LOGGER.info("length of answers is %s", len_of_answer_list)
+        if len_of_answer_list > 0:
```
de-code: minor: I believe if you inline `len` to `if len(list_of_answers_of_one_survey) > 0:` then `pylint` will complain.

That is because the more `pythonic` way is to just use `if list_of_answers_of_one_survey:`

dags/surveymonkey_data_import_pipeline.py
```diff
@@ -117,6 +119,39 @@ def surveymonkey_survey_questions_etl(**kwargs):
...
+                    get_bq_json_for_survey_answers_response_json(answers_dict)
+                ]
+                load_given_json_list_data_from_tempdir_to_bq(
+                    project_name=data_config.project_name,
+                    dataset_name=data_config.dataset_name,
+                    table_name=data_config.survey_answers_table_name,
+                    json_list=survey_answers_list
+                )
+        else:
+            LOGGER.info("There is no answer for the survey %s", survey_id)
```
de-code: minor: to reduce nesting, one common practise is to use early stopping, e.g.:

```python
if not list_of_answers_of_one_survey:
    LOGGER.info("There is no answer for the survey %s", survey_id)
    continue
for answers_dict in list_of_answers_of_one_survey:
    # ...
```

Although in this case the `continue` may not even be necessary if we are just "looping" over an empty list.

dags/surveymonkey_data_import_pipeline.py
```diff
@@ -117,6 +119,39 @@ def surveymonkey_survey_questions_etl(**kwargs):
...
+            survey_id=survey_id
+        )
+        len_of_answer_list = len(list_of_answers_of_one_survey)
+        LOGGER.info("length of answers is %s", len_of_answer_list)
+        if len_of_answer_list > 0:
+            for answers_dict in list_of_answers_of_one_survey:
+                survey_answers_list = [
+                    get_bq_json_for_survey_answers_response_json(answers_dict)
+                ]
+                load_given_json_list_data_from_tempdir_to_bq(
```
de-code: comment; perhaps it would be worth considering loading answers of all surveys in one go?

de-code: actually looking at this again, this is related to a single survey only, isn't it?
In this case it definitely seems to be worth to load all answers of a survey in one go.

data_pipeline/surveymonkey/surveymonkey_etl.py
```diff
@@ -54,3 +54,43 @@ def get_bq_json_for_survey_questions_response_json(
...
+
+
+def get_bq_json_for_survey_answers_response_json(
+    survey_response_json: dict
+) -> dict:
+    return {
+        "survey_answer_id": survey_response_json["id"],
+        "survey_id": survey_response_json["survey_id"],
+        "response_status": survey_response_json["response_status"],
+        "total_time_spent": survey_response_json["total_time"],
```
de-code: I would add the unit, e.g. `total_time_spent_in_secs`

data_pipeline/surveymonkey/surveymonkey_etl.py
```diff
@@ -54,3 +54,43 @@ def get_bq_json_for_survey_questions_response_json(
...
+        "date_modified": survey_response_json["date_modified"],
```
de-code: The naming convention that I tried to follow is to put the type at the end, e.g. `modified_date` if it is a date, or `modified_timestamp` if it is actually a timestamp.

dags/surveymonkey_data_import_pipeline.py
HazalCiplak: Actually I used a variable because I put the length in logging too, so not to call it two times, if it make sense?

data_pipeline/surveymonkey/surveymonkey_etl.py
HazalCiplak: Yes, `modified_timestamp` would be better, I am changing it.

dags/surveymonkey_data_import_pipeline.py
HazalCiplak: I have changed it as I don't need to call one more time :)

dags/surveymonkey_data_import_pipeline.py
HazalCiplak: Yes it is related to just one survey, we can append all answers in same list before loading but in that case it can be a huge list.

de-code: Some thoughts on potential benefits:

- It will be semi-atomic: Either we imported the answers of a survey or we haven't (there is no chance of just having imported the first five answers).
- It should be a lot more efficient
- It may even help with schema detection as it can look at more answers

But maybe I don't fully appreciate how huge the list can get. What sort of size are we talking about?

HazalCiplak: Actually size of the list will be updated after a meeting with Mael, maybe we will eliminate some of the surveys to get their data. I will change it as a one-list to load.

dags/surveymonkey_data_import_pipeline.py
```diff
@@ -117,6 +119,37 @@ def surveymonkey_survey_questions_etl(**kwargs):
...
+        project_name=data_config.project_name,
+        dataset_name=data_config.dataset_name,
+        table_name_source=data_config.survey_list_table_name,
+        column_name=data_config.survey_id_column_name
+    ).values.tolist()
+
+    for survey_id_list in survey_id_list_from_df:
+        survey_id = str(survey_id_list[1])
+        LOGGER.info("answers for survey_id: %s", survey_id)
+        list_of_answers_of_one_survey = get_survey_answer_details(
```
de-code: I think what threw me off a bit, which may be obvious when thinking more about it, is that this is actually the list of answers for each user. Correct? (Or survey responses)
For me intuitively, an answer is a single answer by a single user to a single question.
A survey response I would think of from a single user, could be a collection of answers or a single answer.
i.e. perhaps the naming could be slightly clearer?

dags/surveymonkey_data_import_pipeline.py
de-code: How will the size of the list change after the meeting?
In any case I think I got confused with the answer response as well.
Importing the responses probably doesn't need to be atomic, because in "normal" operation we may start importing data before the survey is closed.
But calling the bigquery import for every response individually seem to be too frequent still (e.g. 2000 times).
So at least batching could be useful.

dags/surveymonkey_data_import_pipeline.py
HazalCiplak: It is actually collection of all answers for one survey. maybe I can change the name `list_of_all_answers_for_one_survey` to make it more clear. There is **no user breakdown** for surveys in any part of the code.

data_pipeline/surveymonkey/surveymonkey_etl.py
```diff
@@ -56,14 +57,27 @@ def get_bq_json_for_survey_questions_response_json(
...
+    access_token: str,
+    page_url: str
+) -> dict:
+    headers = get_surveymonkey_api_headers(access_token)
+    return requests.get(page_url, headers=headers).json()
+
+
+def iter_survey_answer_detail_response(access_token: str, survey_id: str) -> Iterable[dict]:
     headers = get_surveymonkey_api_headers(access_token)
     response = requests.get(
```
de-code: I guess this could now use `get_survey_answers_in_next_page`? (maybe with another name)

data_pipeline/surveymonkey/surveymonkey_etl.py
```diff
@@ -56,14 +57,27 @@ def get_bq_json_for_survey_questions_response_json(
...
-        f'https://api.surveymonkey.com/v3/surveys/{survey_id}/responses/bulk',
+        f'https://api.surveymonkey.com/v3/surveys/{survey_id}/responses/bulk/?per_page=100',
```
de-code: Just noticing that SurveyMonkey seem to be the using the term `responses` or `response` in the URL (instead of `answers`).

data_pipeline/utils/data_store/bq_data_service.py
```diff
@@ -466,7 +466,7 @@ def load_given_json_list_data_from_tempdir_to_bq(
     project_name: str,
     dataset_name: str,
     table_name: str,
-    json_list: str
+    json_list: list
```
de-code: actually it should accept an iterable

dags/surveymonkey_data_import_pipeline.py
```diff
@@ -131,23 +130,19 @@ def surveymonkey_survey_answers_etl(**kwargs):
...
         survey_id = str(survey_id_list[1])
         LOGGER.info("answers for survey_id: %s", survey_id)
-        list_of_answers_of_one_survey = get_survey_answer_details(
+        iterable_of_answers_of_one_survey = iter_formated_survey_user_answers(
             access_token=get_surveymonkey_access_token(),
             survey_id=survey_id
         )
-        LOGGER.info("length of answers is %s", len(list_of_answers_of_one_survey))
-        if not list_of_answers_of_one_survey:
+        if not iterable_of_answers_of_one_survey:
```
de-code: I think this won't work with an iterable (because you need to iterate over the iterable in order to know whether it is empty).
Maybe in this case we could just remove the logging?

data_pipeline/utils/data_store/bq_data_service.py
de-code: In practice this will usually be `Iterable[dict]`. But in theory it could have other types. So maybe `Iterable[Any]` would be more accurate.

de-code (approved):

data_pipeline/surveymonkey/surveymonkey_etl.py
HazalCiplak: yes but to avoid the confusion between api response and survey response, decided to use answers for survey responses as there is no difference in meanings.

data_pipeline/surveymonkey/surveymonkey_etl.py
HazalCiplak: Sorry I am not sure if I understand the point here.

data_pipeline/utils/data_store/bq_data_service.py
HazalCiplak: OK!

data_pipeline/surveymonkey/surveymonkey_etl.py
de-code: I mean this seem to be doing exactly the same as `get_survey_answers_in_next_page`. Just that you'd be passing in a different `page_url`.

de-code: i.e.:

```python
response_json = get_survey_answers_in_next_page(
    access_token,
    f'https://api.surveymonkey.com/v3/surveys/{survey_id}/responses/bulk/?per_page=100'
)
```

But then `get_survey_answers_in_next_page` could be renamed to something like `get_survey_answers_page_response`.

HazalCiplak: I see. I have changed the function name to `iter_survey_answers`. I didn’t add `next page` as it is not getting data for only next page. Hope it seem more clear now.

de-code: Maybe a confusion. You already have the function `get_survey_answers_in_next_page` which you call further down. My point is that you could call it here as well. At that time `get_survey_answers_in_next_page` could do with a different name because it isn't specific to "next page".

HazalCiplak: Actually the function has been created specifically getting data for the `next page`, otherwise we are using survey_id to request the responses. 

```python
def iter_survey_answers(access_token: str, survey_id: str) -> Iterable[dict]:
    headers = get_surveymonkey_api_headers(access_token)
    response = requests.get(
        f'https://api.surveymonkey.com/v3/surveys/{survey_id}/responses/bulk/?per_page=100',
        headers=headers
    )
    response_json = response.json()
    yield from response_json["data"] # THIS PART IS DIRECT RESPONSE
    while "next" in response_json["links"]: # IF THERE IS NEXT PAGE LINK, THEN WE ARE USING THIS FUNCTION
        response_json = get_survey_answers_in_next_page(access_token, response_json["links"]["next"])
        yield from response_json["data"]
```

de-code: Maybe a quick call? I think we are talking past each other :-)

PR #789, opened 2021-12-01
SurveyMonkey question answers option parse

HazalCiplak:
https://github.com/elifesciences/data-hub-issues/issues/284

### DIFF
--- a/data_pipeline/surveymonkey/surveymonkey_etl.py
+++ b/data_pipeline/surveymonkey/surveymonkey_etl.py
@@ -38,34 +38,31 @@ def get_survey_question_details(access_token: str, survey_id: str) -> list:
 
 
 def get_no_answer_options_for_question_json() -> dict:
+    """Added for consistency of the schema in BQ"""
     return {
-        "id": None,
-        "text": None,
-        "type": None,
-        "weight": None
+        "id": "",
+        "text": "",
+        "type": "",
+        "weight": ""
     }
 
 
 def parse_answer_options_in_question_answer_json(
-    question_answer_response_json: dict,
-    option_key: str
+    question_answer_response_json: dict
 ) -> dict:
-    if isinstance(question_answer_response_json[option_key], list):
-        result_list = []
-        for answer_option in question_answer_response_json[option_key]:
-            result_list.append({
-                "id": answer_option.get("id"),
-                "text": answer_option.get("text"),
-                "type": answer_option.get("type"),
-                "weight": answer_option.get("weight")
-            })
-        return result_list
-    return [{
-        "id": question_answer_response_json[option_key].get("id"),
-        "text": question_answer_response_json[option_key].get("text"),
-        "type": question_answer_response_json[option_key].get("type"),
-        "weight": question_answer_response_json[option_key].get("weight")
-    }]
+    if isinstance(question_answer_response_json, list):
+        answer_options = question_answer_response_json
+    else:
+        answer_options = [question_answer_response_json]
+    return [
+        {
+            "id": str(answer_option.get("id", "")),
+            "text": str(answer_option.get("text", "")),
+            "type": str(answer_option.get("type", "")),
+            "weight": str(answer_option.get("weight", ""))
+        }
+        for answer_option in answer_options
+    ]
 
 
 def parse_answers_in_question_json(question_response_json: dict) -> dict:
@@ -76,22 +73,11 @@ def parse_answers_in_question_json(question_response_json: dict) -> dict:
         "cols": [get_no_answer_options_for_question_json()]
     }
     if "answers" in question_response_json:
-        if "choices" in question_response_json["answers"]:
-            result["choices"] = parse_answer_options_in_question_answer_json(
-                question_response_json["answers"],
-                "choices")
-        if "other" in question_response_json["answers"]:
-            result["other"] = parse_answer_options_in_question_answer_json(
-                question_response_json["answers"],
-                "other")
-        if "rows" in question_response_json["answers"]:
-            result["rows"] = parse_answer_options_in_question_answer_json(
-                question_response_json["answers"],
-                "rows")
-        if "cols" in question_response_json["answers"]:
-            result["cols"] = parse_answer_options_in_question_answer_json(
-                question_response_json["answers"],
-                "cols")
+        for key in ["choices", "other", "rows", "cols"]:
+            if key in question_response_json["answers"]:
+                result[key] = parse_answer_options_in_question_answer_json(
+                    question_response_json["answers"][key]
+                )
     return result
 
 
@@ -131,7 +117,7 @@ def iter_survey_answers(access_token: str, survey_id: str) -> Iterable[dict]:
         access_token,
         f"https://api.surveymonkey.com/v3/surveys/{survey_id}/responses/bulk/?per_page=100"
     )
-    if not response_json["data"]:
+    if not response_json.get("data"):
         LOGGER.info("No answers for the survey: %s", survey_id)
     yield from response_json["data"]
     while "next" in response_json["links"]:
--- a/data_pipeline/surveymonkey/surveymonkey_etl.py
+++ b/data_pipeline/surveymonkey/surveymonkey_etl.py
@@ -37,16 +37,6 @@ def get_survey_question_details(access_token: str, survey_id: str) -> list:
     return reponse_json
 
 
-def get_no_answer_options_for_question_json() -> dict:
-    """Added for consistency of the schema in BQ"""
-    return {
-        "id": "",
-        "text": "",
-        "type": "",
-        "weight": ""
-    }
-
-
 def parse_answer_options_in_question_answer_json(
     question_answer_response_json: dict
 ) -> dict:
@@ -56,22 +46,17 @@ def parse_answer_options_in_question_answer_json(
         answer_options = [question_answer_response_json]
     return [
         {
-            "id": str(answer_option.get("id", "")),
-            "text": str(answer_option.get("text", "")),
-            "type": str(answer_option.get("type", "")),
-            "weight": str(answer_option.get("weight", ""))
+            "id": answer_option.get("id", ""),
+            "text": answer_option.get("text", ""),
+            "type": answer_option.get("type", ""),
+            "weight": answer_option.get("weight", -99)
         }
         for answer_option in answer_options
     ]
 
 
 def parse_answers_in_question_json(question_response_json: dict) -> dict:
-    result = {
-        "choices": [get_no_answer_options_for_question_json()],
-        "other": [get_no_answer_options_for_question_json()],
-        "rows": [get_no_answer_options_for_question_json()],
-        "cols": [get_no_answer_options_for_question_json()]
-    }
+    result = {}
     if "answers" in question_response_json:
         for key in ["choices", "other", "rows", "cols"]:
             if key in question_response_json["answers"]:
--- a/data_pipeline/surveymonkey/surveymonkey_etl.py
+++ b/data_pipeline/surveymonkey/surveymonkey_etl.py
@@ -37,7 +37,7 @@ def get_survey_question_details(access_token: str, survey_id: str) -> list:
     return reponse_json
 
 
-def get_no_answer_options_from_question_json() -> dict:
+def get_no_answer_options_for_question_json() -> dict:
     return {
         "id": None,
         "text": None,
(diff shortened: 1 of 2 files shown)


tests/unit_test/surveymonkey/surveymonkey_etl_test.py
```diff
@@ -55,20 +55,188 @@ def test_should_ignore_other_field(self):
...
+        })
+        assert result["questions"] == [
+            {
+                "question_id": "Q_ID",
+                "question_title": "Is this the question?",
+                "question_type": "matrix",
+                "question_subtype": "rating",
+                "question_answers": [{
+                    "choices": [{
+                        "id": None,
```
de-code: Wouldn't we in this case expect `choices` to be an empty list or not set?

tests/unit_test/surveymonkey/surveymonkey_etl_test.py
```diff
@@ -55,20 +55,188 @@ def test_should_ignore_other_field(self):
...
+                        "type": None,
+                        "weight": None
                     }]
                 }]
+            }
+        ]
+
+    def test_should_extract_question_answers(self):
+        result = get_bq_json_for_survey_questions_response_json({
+            **DEFAULT_SURVEY_QUESTIONS_RESPONSE_JSON,
```
de-code, line 84: It might be useful to explicitly set the values here. Otherwise reading this test it is not clear why we should for example expect there to be a choice with the text `This is the choice`.
Alternatively, if the test is rather about to test just that all of the elements are extracted without testing the specifics, then you could consider moving the expected result to a constant close to the response json constant.
You could then add more specific tests for `parse_answer_options_in_question_answer_json`.

data_pipeline/surveymonkey/surveymonkey_etl.py
```diff
@@ -30,13 +30,71 @@ def get_survey_list(access_token: str) -> list:
...
+        "id": None,
+        "text": None,
+        "type": None,
+        "weight": None
+    }
+
+
+def parse_answer_options_in_question_answer_json(
+    question_answer_response_json: dict,
+    option_key: str
```
de-code: This function doesn't seem to be using `option_key` apart from using it as a key for `question_answer_response_json`. Maybe in that case it would make sense to just pass `question_answer_response_json[option_key]` to the function?

data_pipeline/surveymonkey/surveymonkey_etl.py
```diff
@@ -30,13 +30,71 @@ def get_survey_list(access_token: str) -> list:
...
+) -> dict:
+    if isinstance(question_answer_response_json[option_key], list):
```
de-code: Maybe if the value is either a `dict` or `List[dict]` (you could declare it as the type `Union[dict, List[dict]]`),
and  the dict in both cases have the same meaning - just hat it is wrapped in list,
then you could just wrap the value in a list if it isn't, and use the same code path?

e.g.

```python
if isinstance(question_answer_response_json[option_key], list):
    answer_options = question_answer_response_json[option_key]
else:
    answer_options = [question_answer_response_json[option_key]]
return [
    {
        "id": answer_option.get("id"),
        "text": answer_option.get("text"),
        "type": answer_option.get("type"),
        "weight": answer_option.get("weight")
    }
    for answer_option in answer_options
]
```

data_pipeline/surveymonkey/surveymonkey_etl.py
```diff
@@ -30,13 +30,71 @@ def get_survey_list(access_token: str) -> list:
...
+
+def parse_answers_in_question_json(question_response_json: dict) -> dict:
+    result = {
+        "choices": [get_no_answer_options_for_question_json()],
+        "other": [get_no_answer_options_for_question_json()],
+        "rows": [get_no_answer_options_for_question_json()],
+        "cols": [get_no_answer_options_for_question_json()]
+    }
+    if "answers" in question_response_json:
+        if "choices" in question_response_json["answers"]:
```
de-code: I don't know much about the survey response, but if you can use `parse_answer_options_in_question_answer_json` for all of the answers, it seems like here you could iterate over a list of keys to reduce duplication?
e.g.

```python
for key in ["choices", "other", "rows", "cols"]:
    result[key] = parse_answer_options_in_question_answer_json(
        question_response_json["answers"],
        key
    )
```

data_pipeline/surveymonkey/surveymonkey_etl.py
```diff
@@ -30,13 +30,71 @@ def get_survey_list(access_token: str) -> list:
...
+        "id": question_answer_response_json[option_key].get("id"),
+        "text": question_answer_response_json[option_key].get("text"),
+        "type": question_answer_response_json[option_key].get("type"),
+        "weight": question_answer_response_json[option_key].get("weight")
+    }]
+
+
+def parse_answers_in_question_json(question_response_json: dict) -> dict:
+    result = {
+        "choices": [get_no_answer_options_for_question_json()],
```
de-code: I see now where the "no answer" value comes from. I don't know the benefit of having the "empty" option in this case. Maybe it is worth adding a comment regarding why we add it.

Maybe I would suggest to move that logic to `parse_answer_options_in_question_answer_json`.
Then you could potentially also remove the `if` statements before calling `parse_answer_options_in_question_answer_json`?

data_pipeline/surveymonkey/surveymonkey_etl.py
HazalCiplak: Thank you I will use this!

data_pipeline/surveymonkey/surveymonkey_etl.py
HazalCiplak: Added:

```python
    for key in ["choices", "other", "rows", "cols"]:
            if key in question_response_json["answers"]:
                result[key] = parse_answer_options_in_question_answer_json(
                    question_response_json["answers"][key]
                )
```

tests/unit_test/surveymonkey/surveymonkey_etl_test.py
```diff
@@ -128,7 +128,7 @@ def test_should_extract_question_answers(self):
                         "rows": [{
                             "id": "ROW_ID",
                             "text": "This is the row",
-                            "weight": 0
+                            "weight": "3"
```
de-code: This is my recommendation:
I mentioned it yesterday, I don't think we should convert numbers to a string. It basically changes is from a continuous value that one can average etc, to a categorical value.
If there are inconsistencies, it would be good to document them (e.g. in tests).
Alternatively just leave `weight` out until needed and we had more time at looking at various examples.

data_pipeline/surveymonkey/surveymonkey_etl.py
```diff
@@ -38,34 +38,31 @@ def get_survey_question_details(access_token: str, survey_id: str) -> list:
 
 
 def get_no_answer_options_for_question_json() -> dict:
+    """Added for consistency of the schema in BQ"""
```
de-code: My preference:
I would really prefer if we didn't have a "no answer" option unless it is needed for the end result.
e.g. if we have a subject area filter and need an option saying "No Subject Area".

In this case it doesn't seem to match that.
And it if you counted the "options" it would say `1` even though it should be `0`.

Just adding it because it gets past schema validation errors does not seem to be a good reason to add it.
It would be better to look at the schema validation errors.
Why can't for example `choices` be an empty list or optional?
What is the schema that is being generated?

data_pipeline/surveymonkey/surveymonkey_etl.py
```diff
@@ -38,34 +38,31 @@ def get_survey_question_details(access_token: str, survey_id: str) -> list:
...
+    if isinstance(question_answer_response_json, list):
+        answer_options = question_answer_response_json
+    else:
+        answer_options = [question_answer_response_json]
+    return [
+        {
+            "id": str(answer_option.get("id", "")),
+            "text": str(answer_option.get("text", "")),
+            "type": str(answer_option.get("type", "")),
+            "weight": str(answer_option.get("weight", ""))
```
de-code: In addition to the comment regarding converting weight to a str. In general while converting to str works, it would mask errors. It would be good to have actual examples where the types are inconsistent. e.g. in what case is `weight` an int or another type? Same for the other fields.
It would be better if it fails so we can look at it.

data_pipeline/surveymonkey/surveymonkey_etl.py
```diff
@@ -38,34 +38,31 @@ def get_survey_question_details(access_token: str, survey_id: str) -> list:
...
-        "type": question_answer_response_json[option_key].get("type"),
-        "weight": question_answer_response_json[option_key].get("weight")
-    }]
+    if isinstance(question_answer_response_json, list):
+        answer_options = question_answer_response_json
+    else:
+        answer_options = [question_answer_response_json]
+    return [
+        {
+            "id": str(answer_option.get("id", "")),
```
de-code: Is `id` actually optional?

tests/unit_test/surveymonkey/surveymonkey_etl_test.py
HazalCiplak: OK

data_pipeline/surveymonkey/surveymonkey_etl.py
HazalCiplak: OK

data_pipeline/surveymonkey/surveymonkey_etl.py
HazalCiplak: OK

data_pipeline/surveymonkey/surveymonkey_etl.py
HazalCiplak: If the answer is just text in that case there is no id for it.

de-code: That is something that would be great to capture in a test.
e.g. you could test `parse_answer_options_in_question_answer_json` specifically and have a test along the lines of `test_should_not_return_id_for_text_only_response` (or something like that).

tests/unit_test/surveymonkey/surveymonkey_etl_test.py
HazalCiplak: `None` is causing problem (while creating schema and loading the data in BQ) if I don't add a default value in the `get` function? What can I add here as default value, for string values I was adding `""`. Or do you have any idea how can I manage the None value in json not to cause a problem.

```python
"id": answer_option.get("id",""),
"text": answer_option.get("text",""),
"type": answer_option.get("type",""),
"weight": answer_option.get("weight")
```

data_pipeline/surveymonkey/surveymonkey_etl.py
HazalCiplak: `None` is causing problem (while creating schema and loading the data in BQ) if I don't add a default value in the `get` function? What can I add here as default value, for string values I was adding `""`. Or do you have any idea how can I manage the None value in json not to cause a problem.

```python
"id": answer_option.get("id",""),
"text": answer_option.get("text",""),
"type": answer_option.get("type",""),
"weight": answer_option.get("weight")
```

data_pipeline/surveymonkey/surveymonkey_etl.py
```diff
@@ -48,7 +77,10 @@ def get_bq_json_for_survey_questions_response_json(
         "questions": [
             {
                 "question_id": question_response_json["id"],
-                "question_title": question_response_json["headings"][0]["heading"]
+                "question_title": question_response_json["headings"][0]["heading"],
+                "question_type": question_response_json["family"],
+                "question_subtype": question_response_json["subtype"],
+                "question_answers": [parse_answers_in_question_json(question_response_json)]
```
de-code: I guess we don't need to wrap it into an array? It seems we are using an array, which itself contains multiple arrays.

data_pipeline/surveymonkey/surveymonkey_etl.py
```diff
@@ -30,13 +30,42 @@ def get_survey_list(access_token: str) -> list:
...
+    if isinstance(question_answer_response_json, list):
+        answer_options = question_answer_response_json
+    else:
+        answer_options = [question_answer_response_json]
+    return [
+        {
+            "id": answer_option.get("id", ""),
+            "text": answer_option.get("text", ""),
+            "type": answer_option.get("type", ""),
+            "weight": answer_option.get("weight", -99)
```
de-code: Do we need the default value? Can't weight be None? Or don't include weight in the resulting object?
As this is a magic number that we would have to encode in SQL and would cause issues with aggregation functions if not handled specifically.

data_pipeline/surveymonkey/surveymonkey_etl.py
```diff
@@ -30,13 +30,42 @@ def get_survey_list(access_token: str) -> list:
...
-        f'https://api.surveymonkey.com/v3/surveys/{survey_id}/details',
+        f"https://api.surveymonkey.com/v3/surveys/{survey_id}/details",
         headers=headers
     )
     reponse_json = response.json()
     return reponse_json
 
 
+def parse_answer_options_in_question_answer_json(
+    question_answer_response_json: dict
```
de-code, line 42: The type should now be: `Union[dict, List[dict]]`

tests/unit_test/surveymonkey/surveymonkey_etl_test.py
```diff
@@ -55,20 +55,181 @@ def test_should_ignore_other_field(self):
...
         assert result["questions"] == [
-            {"question_id": "Q_ID", "question_title": "Is this the question?"}
+            {
+                "question_id": "Q_ID",
+                "question_title": "Is this the question?",
+                "question_type": "matrix",
+                "question_subtype": "rating",
+                "question_answers": [{
+                    "other": [{
+                            "id": "",
```
de-code: minor indentation

data_pipeline/surveymonkey/surveymonkey_etl.py
de-code: I see your response now here. In general BigQuery fields can be optional so that shouldn't be an issue in general.
I think it would be useful to see what schema is being generated that is causing the issue, and what the exact error is.
It could be that the schema generation or JSON output doesn't work well with `None` fields. And it may work better if the field isn't there at all.

data_pipeline/surveymonkey/surveymonkey_etl.py
de-code: (This links to https://github.com/elifesciences/data-hub-core-airflow-dags/pull/789#discussion_r761809544 which is now shown as outdated)

data_pipeline/surveymonkey/surveymonkey_etl.py
HazalCiplak: Will try it to see the result.

data_pipeline/surveymonkey/surveymonkey_etl.py
HazalCiplak: Without default value I get creating schema problems, maybe we can look into thin in data catch-up

data_pipeline/surveymonkey/surveymonkey_etl.py
de-code: I gave it a quick test with some dummy data:

```python
import json
from io import StringIO
from bigquery_schema_generator.generate_schema import SchemaGenerator

generator = SchemaGenerator(
    input_format="json",
    quoted_values_are_strings=True
)
json_data = {"data": [{"text": "value1", "weight": 1}, {"text": "value1", "weight": None}]}
json_str = json.dumps(json_data)
print('json: %r' % json_str)
schema_map, _ = generator.deduce_schema(StringIO(json_str))
schema = generator.flatten_schema(schema_map)
print('schema: %s' % schema)
```

Output:

```text
json: '{"data": [{"text": "value1", "weight": 1}, {"text": "value1", "weight": null}]}'
schema: [OrderedDict([('fields', [OrderedDict([('mode', 'NULLABLE'), ('name', 'text'), ('type', 'STRING')]), OrderedDict([('mode', 'NULLABLE'), ('name', 'weight'), ('type', 'INTEGER')])]), ('mode', 'REPEATED'), ('name', 'data'), ('type', 'RECORD')])]
```

This shows that in this case `None` is serialized as `null`.
It doesn't seem to trip over the schema generator. It is actually declaring both `text` and `weight` as `NULLABLE`. Which is probably good as it allows us to deprecate fields over time.

data_pipeline/surveymonkey/surveymonkey_etl.py
de-code: Just adding a comment that this has now changed to use an empty list instead.

data_pipeline/surveymonkey/surveymonkey_etl.py
de-code: Just adding a comment that it seems that it was the BigQuery import that was failing with `null` values.
We changed the code to not include None fields. i.e. the answer option dictionary will only have non-none values.
As a result, the resulting BigQuery schema may only show a subset of the four fields, depending on the question type / parent answer field (e.g. `choices` contains `weight`, but others might not)

de-code (approved):

PR #856, opened 2022-02-03
civiCRM email summary report etl

HazalCiplak:
https://github.com/elifesciences/data-hub-issues/issues/332

### DIFF
--- a/dags/civicrm_email_report_import_pipeline.py
+++ b/dags/civicrm_email_report_import_pipeline.py
@@ -1,9 +1,10 @@
 # Note: DagBag.process_file skips files without "airflow" or "DAG" in them
 
 import os
-from ruamel import yaml
 import logging
 from datetime import timedelta
+import base64
+from ruamel import yaml
 
 from data_pipeline.utils.dags.data_pipeline_dag_utils import create_dag, create_python_task
 from data_pipeline.utils.pipeline_file_io import get_yaml_file_as_dict
@@ -29,6 +30,7 @@ DAG_ID = "Civicrm_Email_Summary_Report_Pipeline"
 CIVICRM_EMAIL_REPORT_DATA_PIPELINE_SCHEDULE_INTERVAL_ENV_NAME = (
     "CIVICRM_EMAIL_REPORT_DATA_PIPELINE_SCHEDULE_INTERVAL"
 )
+
 
 def get_env_var_or_use_default(env_var_name, default_value=None):
     return os.getenv(env_var_name, default_value)
@@ -63,10 +65,11 @@ def data_config_from_xcom(context):
 
 def get_civicrm_credential(key_name: str):
     secret_file = get_env_var_or_use_default(CIVICRM_KEY_SECRET_FILE_PATH_ENV, "")
+    # pylint: disable=consider-using-with
     data = yaml.safe_load(open(secret_file))
-    if key_name=="api_key":
-        return data["data"]["civi_api_key.txt"]
-    return data["data"]["civi_site_key.txt"]
+    if key_name == "api_key":
+        return base64.b64decode(data["data"]["civi_api_key.txt"])
+    return base64.b64decode(data["data"]["civi_site_key.txt"])
 
 
 def civicrm_email_report_etl(**kwargs):
--- a/dags/civicrm_email_report_import_pipeline.py
+++ b/dags/civicrm_email_report_import_pipeline.py
@@ -8,7 +8,7 @@ from ruamel import yaml
 
 from data_pipeline.utils.dags.data_pipeline_dag_utils import create_dag, create_python_task
 from data_pipeline.utils.pipeline_file_io import get_yaml_file_as_dict
-from data_pipeline.civicrm_email_report.civicrm_email_report import get_email_report
+from data_pipeline.civicrm_email_report.civicrm_email_report import iter_email_report
 from data_pipeline.utils.data_store.bq_data_service import (
     get_distinct_values_from_bq,
     load_given_json_list_data_from_tempdir_to_bq
@@ -80,22 +80,20 @@ def civicrm_email_report_etl(**kwargs):
         table_name_source=data_config.email_id_source_table,
         column_name=data_config.email_id_column
     ).values.tolist()
-
-    for email_id in email_id_list:
-        LOGGER.info("mail_id to process is: %s", int(email_id[1]))
-        email_report = get_email_report(
-            url=data_config.civicrm_api_url,
-            mail_id=int(email_id[1]),
-            api_key=get_civicrm_credential("api_key"),
-            site_key=get_civicrm_credential("site_key"),
-        )
-
-        load_given_json_list_data_from_tempdir_to_bq(
-            project_name=data_config.project_name,
-            dataset_name=data_config.dataset_name,
-            table_name=data_config.table_name,
-            json_list=[email_report]
-        )
+    print(email_id_list)
+    
+    email_reports = iter_email_report(
+        url=data_config.civicrm_api_url,
+        mail_id_list=email_id_list,
+        api_key=get_civicrm_credential("api_key"),
+        site_key=get_civicrm_credential("site_key"),
+    )
+    load_given_json_list_data_from_tempdir_to_bq(
+        project_name=data_config.project_name,
+        dataset_name=data_config.dataset_name,
+        table_name=data_config.table_name,
+        json_list=email_reports
+    )
 
 
 CIVICRM_EMAIL_DAG = create_dag(
--- a/dags/civicrm_email_report_import_pipeline.py
+++ b/dags/civicrm_email_report_import_pipeline.py
@@ -64,9 +64,7 @@ def data_config_from_xcom(context):
 
 
 def get_civicrm_credential(env_var_name: str):
-    return read_file_content(
-            get_env_var_or_use_default(env_var_name, "")
-        )
+    return read_file_content(get_env_var_or_use_default(env_var_name, ""))
 
 
 def civicrm_email_report_etl(**kwargs):
--- a/dags/civicrm_email_report_import_pipeline.py
+++ b/dags/civicrm_email_report_import_pipeline.py
@@ -93,7 +93,7 @@ def civicrm_email_report_etl(**kwargs):
         load_given_json_list_data_from_tempdir_to_bq(
             project_name=data_config.project_name,
             dataset_name=data_config.dataset_name,
-            table_name=data_config.email_id_source_table,
+            table_name=data_config.table_name,
             json_list=[email_report]
         )
 
--- a/dags/civicrm_email_report_import_pipeline.py
+++ b/dags/civicrm_email_report_import_pipeline.py
@@ -3,8 +3,6 @@
 import os
 import logging
 from datetime import timedelta
-import base64
-from ruamel import yaml
 
 from data_pipeline.utils.dags.data_pipeline_dag_utils import create_dag, create_python_task
 from data_pipeline.utils.pipeline_file_io import get_yaml_file_as_dict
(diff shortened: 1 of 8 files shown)


data_pipeline/civicrm_email_report/civicrm_email_report.py
```diff
@@ -0,0 +1,61 @@
...
+
+def get_email_report(
+    url: str,
+    mail_id: int,
+    api_key: str,
+    site_key: str
+) -> dict:
+    response = requests.post(url=url, data=get_connection_parameters(mail_id, api_key, site_key))
+    response.raise_for_status()
+    dict_response = eval(response.text)
```
de-code: I'd always be careful with using `eval` on anything as it could run any code. Why do we use it here?

HazalCiplak: To convert the string to a dict

HazalCiplak:
Hi @de-code, I am not pretty sure about the authentication key handling in the code, it would be good to look at it together when you find time today or tomorrow before deploying.

data_pipeline/civicrm_email_report/civicrm_email_report.py
HazalCiplak: I found another way to doing it. What about this one?
```python
>>> import ast
>>> d = ast.literal_eval("{'key1' : 'value1', 'key2' : 'value2'}")
>>> print(type(d))
<class 'dict'>
```

```
literal_eval(node_or_string)
    Safely evaluate an expression node or a string containing a Python
    expression.  The string or node provided may only consist of the following
    Python literal structures: strings, numbers, tuples, lists, dicts, booleans,
    and None.
```

or `json.loads()`

HazalCiplak: used json.loads()

data_pipeline/civicrm_email_report/civicrm_email_report.py
```diff
@@ -0,0 +1,65 @@
...
+
+def get_email_report(
+    url: str,
+    mail_id: int,
+    api_key: str,
+    site_key: str
+) -> dict:
+    response = requests.post(url=url, data=get_connection_parameters(mail_id, api_key, site_key))
+    response.raise_for_status()
+    dict_response = json.loads(response.text)
```
de-code: yes, using the `json` library is much better. I believe the response should also have a `json` function?

data_pipeline/civicrm_email_report/civicrm_email_report.py
```diff
@@ -0,0 +1,65 @@
...
+        "clickthrough_rate": float(
+            dict_response["values"][mail_id_str]["clickthrough_rate"].replace("%", "")
+        ),
+        "imported_timestamp": get_current_timestamp()
+    }
+
+
+def iter_email_report(
+    url: str,
+    mail_id_list: list,
```
de-code: I would maybe add a more specific type. e.g. `Sequence[str]`.

Currently it seems that we are passing a list of lists / tuples, whereas the mail id as at index `1`.
Could we just pass in the list of mail ids?

data_pipeline/civicrm_email_report/civicrm_email_report.py
HazalCiplak: Yes normally I am using it too bu unfortunately it did not work for this case. json function returning something totally different like` <bound method Response.json of <Response [200]>>`

data_pipeline/civicrm_email_report/civicrm_email_report.py
HazalCiplak: We will need to create different function to query it from BQ. But I prefer to use existing one and we are using this function in different pipelines too. 

What is `Sequence[str]` for? `mail_id_list` type `list[list]`?

requirements.txt
```diff
@@ -17,3 +17,4 @@ PyYAML>=5.4
 s3fs==2021.5.0
 six==1.16.0
 oauth2client==4.1.3
+ruamel.yaml==0.17.20
```
de-code: I had always used pyyaml so I was looking into the difference..

Some links that I found:
- https://stackoverflow.com/a/36760452/8676953
- https://github.com/yaml/pyyaml/issues/46

So it seems some of the major advantages is YAML 1.2 and being developed more actively.

dags/civicrm_email_report_import_pipeline.py
```diff
@@ -0,0 +1,122 @@
...
+    data_configuration = CiviCrmEmailReportDataConfig(
+        data_config_dict, deployment_env)
+    LOGGER.info('data_config: %r', data_configuration)
+    return data_configuration
+
+
+def get_civicrm_credential(key_name: str):
+    secret_file = get_env_var_or_use_default(CIVICRM_KEY_SECRET_FILE_PATH_ENV, "")
+    # pylint: disable=consider-using-with
+    data = yaml.safe_load(open(secret_file))
```
de-code: I'd always use a context manger with `open` to ensure the file is closed. e.g.:

```python
with open(secret_file) as fp:
    data = yaml.safe_load(fp)
```

dags/civicrm_email_report_import_pipeline.py
```diff
@@ -0,0 +1,122 @@
...
+        DEFAULT_DEPLOYMENT_ENV
+    )
+    data_configuration = CiviCrmEmailReportDataConfig(
+        data_config_dict, deployment_env)
+    LOGGER.info('data_config: %r', data_configuration)
+    return data_configuration
+
+
+def get_civicrm_credential(key_name: str):
+    secret_file = get_env_var_or_use_default(CIVICRM_KEY_SECRET_FILE_PATH_ENV, "")
```
de-code: I guess by this time, the environment variable is required?

Is that the same configuration that we use for the current Civi API access or is that something new?

data_pipeline/civicrm_email_report/civicrm_email_report.py
de-code: This `<bound method...` would suggest to me that it was the reference to the `json` function rather than the return value.
Maybe you didn't call it?

data_pipeline/civicrm_email_report/civicrm_email_report.py
de-code: Currently it would be `Sequence[Sequence[str]]` or `Sequence[Tuple[str, str]]` or could also be `List[List[str]]`. I don't know what the types of the items are. `Sequence` is often preferable over `List` as an input type as it is less restrictive. We could also accept `Iterable` if we didn't use `len`.

dags/civicrm_email_report_import_pipeline.py
```diff
@@ -0,0 +1,122 @@
...
+    # pylint: disable=consider-using-with
+    data = yaml.safe_load(open(secret_file))
+    if key_name == "api_key":
+        return base64.b64decode(data["data"]["civi_api_key.txt"])
+    return base64.b64decode(data["data"]["civi_site_key.txt"])
+
+
+def civicrm_email_report_etl(**kwargs):
+    data_config = data_config_from_xcom(kwargs)
+    email_id_list = get_distinct_values_from_bq(
```
de-code: What does `get_distinct_values_from_bq` return?
I guess it is a pandas `DataFrame`? What are the columns? You could just select that column.

HazalCiplak: column name can  be `0` probably. Because I did not named it. let me check

HazalCiplak:
```
Data columns (total 2 columns):
 #   Column  Non-Null Count  Dtype
---  ------  --------------  -----
 0   index   3010 non-null   int64
 1   0       3010 non-null   int64
```

data_pipeline/civicrm_email_report/civicrm_email_report.py
HazalCiplak:
```
print(response.json)
print(response.text)

<bound method Response.json of <Response [200]>>
{"is_error":0,"version":3,"count":1,"id":6232,"values":{"6232":{"Delivered":"135","Bounces":"1","Unsubscribers":0,"Unique Clicks":"39","Opened":"251","clickthrough_rate":"28.89%","opened_rate":"185.93%","delivered_rate":"99.26%"}}}
```

HazalCiplak: Probably I misunderstand, what did you mean by not calling? I was thinking the reason is about using `post` request instead of `get` but if I need to call something else I am not aware of it?

data_pipeline/civicrm_email_report/civicrm_email_report.py
HazalCiplak: Now it is just a `list`.

dags/civicrm_email_report_import_pipeline.py
HazalCiplak: used the column name and get the list afterwards.

dags/civicrm_email_report_import_pipeline.py
```diff
@@ -0,0 +1,126 @@
...
+    data_configuration = CiviCrmEmailReportDataConfig(
+        data_config_dict, deployment_env)
+    LOGGER.info('data_config: %r', data_configuration)
+    return data_configuration
+
+
+def get_civicrm_credential(key_name: str):
+    if key_name == "api_key":
+        return read_file_content(
+            get_env_var_or_use_default(CIVICRM_API_KEY_FILE_PATH_ENV, "")
```
de-code: Could we make it so the environment variable is required? It would make error messages more informative.

dags/civicrm_email_report_import_pipeline.py
```diff
@@ -0,0 +1,126 @@
...
+        DEPLOYMENT_ENV_ENV_NAME,
+        DEFAULT_DEPLOYMENT_ENV
+    )
+    data_configuration = CiviCrmEmailReportDataConfig(
+        data_config_dict, deployment_env)
+    LOGGER.info('data_config: %r', data_configuration)
+    return data_configuration
+
+
+def get_civicrm_credential(key_name: str):
```
de-code: minor: Maybe rather than passing in the key name, we could just pass in the env variable name?
Otherwise at the moment it isn't clear what the accepted key names are.

de-code: Or alternatively these could just be two separate functions. They don't really share any code.

data_pipeline/civicrm_email_report/civicrm_email_report.py
```diff
@@ -0,0 +1,74 @@
...
+    mail_id: int,
+    api_key: str,
+    site_key: str
+) -> dict:
+    response = requests.post(url=url, data=get_connection_parameters(mail_id, api_key, site_key))
+    response.raise_for_status()
+    return response.json()
+
+
+def format_email_report(
```
de-code: minor: usually I would associate a `format` function with a text output. Perhaps another name could make it more clear? Maybe `transform`? And maybe it could say from what to what format do we transform it?

dags/civicrm_email_report_import_pipeline.py
```diff
@@ -0,0 +1,126 @@
...
+    load_given_json_list_data_from_tempdir_to_bq
+)
+from data_pipeline.civicrm_email_report.get_civicrm_email_report_data_config import (
+    CiviCrmEmailReportDataConfig
+)
+from data_pipeline.utils.pipeline_file_io import read_file_content
+
+LOGGER = logging.getLogger(__name__)
+
+CIVICRM_KEY_SECRET_FILE_PATH_ENV = "CIVICRM_KEY_SECRET_FILE_PATH"
```
de-code: I believe this can now be removed? (And any other related mention in other files)

dags/civicrm_email_report_import_pipeline.py
HazalCiplak: used env var name

dags/civicrm_email_report_import_pipeline.py
```diff
@@ -0,0 +1,124 @@
...
+        DEFAULT_DEPLOYMENT_ENV
+    )
+    data_configuration = CiviCrmEmailReportDataConfig(
+        data_config_dict, deployment_env)
+    LOGGER.info('data_config: %r', data_configuration)
+    return data_configuration
+
+
+def get_civicrm_credential(env_var_name: str):
+    if env_var_name == "CIVICRM_API_KEY_FILE_PATH_ENV":
```
de-code: You no longer need this condition, now that you pass in the env var name.
You would just use the env var name.

HazalCiplak: You mean creating two different function ? Because I did not get the point of not using condition (if statement)

de-code: Creating two separate functions was the alternative...

Buf if we pass in the environment variable name to the function then we can just use it, e.g.:

```python
    return read_file_content(get_required_env_var(env_var_name))
```

HazalCiplak: Ah yes sorry, my mind is a bit full atm.

PR #865, opened 2022-02-07
tests for civicrm email report 

HazalCiplak:
https://github.com/elifesciences/data-hub-issues/issues/354

### DIFF
--- a/tests/unit_test/civicrm_email/civicrm_email_report_test.py
+++ b/tests/unit_test/civicrm_email/civicrm_email_report_test.py
@@ -56,7 +56,7 @@ class TestTransformEmailReport():
             MAIL_ID
         )
         assert result["opened"] == 251
-    
+
     def test_should_extract_delivered_rate_as_float(self):
         result = transform_email_report(
             DEFAULT_CIVICRM_EMAIL_REPORT_RESPONSE,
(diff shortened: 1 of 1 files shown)


tests/unit_test/civicrm_email/civicrm_email_report_test.py
```diff
@@ -0,0 +1,79 @@
+from data_pipeline.civicrm_email_report.civicrm_email_report import (
+    transform_email_report
+)
+
+MAIL_ID = 100
+
+DEFAULT_CIVICRM_EMAIL_REPORT_RESPONSE = {
+    "values": {
+        str(MAIL_ID): {
```
de-code, line 18: Just curious, the mail id is a string in the response, but an integer in other responses?

HazalCiplak: when it is a key in the response dictionary it is a string. But when we sending it to API we are sending as an integer.

de-code: Interesting. Although thinking about it more, it makes sense. As integers aren't valid JSON keys.

de-code (approved):

PR #866, opened 2022-02-08
changed dependency versions for dask

HazalCiplak:
https://alfred.elifesciences.org/blue/organizations/jenkins/data-hub-airflow-image/detail/data-hub-airflow-image/946/pipeline/

https://github.com/elifesciences/data-hub-issues/issues/357

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,10 +4,10 @@ bigquery-schema-generator==1.5
 bigquery-views-manager>=1.0.3
 boto3==1.17.87
 cattrs==1.10.0
-cloudpickle>=1.5.0
+cloudpickle==1.4.1
 dateparser==1.1.0
-dask[complete]==2021.2.0
-distributed==2021.2.0
+dask[complete]<=2021.2.0, >=2.17.0
+distributed<=2021.2.0, >=2.17.0
 fsspec>=2021.4.0
 google-api-python-client==2.36.0
 google-cloud-bigquery>=1.27.2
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -4,10 +4,10 @@ bigquery-schema-generator==1.5
 bigquery-views-manager>=1.0.3
 boto3==1.17.87
 cattrs==1.10.0
-cloudpickle==1.4.1
+cloudpickle>=1.5.0
 dateparser==1.1.0
-dask[complete]<=2021.2.0
-distributed<=2021.2.0
+dask[complete]>=2022.1.0
```
de-code: I would keep the upper bound as it was, there was probably a reason for it (but not sure), e.g.
`dask[complete]>=2021.1.0,<=2021.2.0`

requirements.txt
```diff
@@ -4,10 +4,10 @@ bigquery-schema-generator==1.5
 bigquery-views-manager>=1.0.3
 boto3==1.17.87
 cattrs==1.10.0
-cloudpickle==1.4.1
+cloudpickle>=1.5.0
```
de-code: cloudpickle caused issues before, I'd try to avoid upgrading it as part of this PR.
Ideally the plan was to just narrow the version range for dask, not actually change the dependencies.

requirements.txt
de-code: Alternatively you could pin the version to `2021.2.0`.
Although I believe the benefit of the range is that it avoids dependency conflicts, within this project and also when installing dependencies for all DAG repos together.
i.e. using a range might still work better overall and doesn't introduce new problems

de-code (approved):

PR #899, opened 2022-02-28
Configured gmail data pipeline config

HazalCiplak:
https://github.com/elifesciences/data-hub-issues/issues/333

### DIFF
--- a/dags/gmail_data_import_controller.py
+++ b/dags/gmail_data_import_controller.py
@@ -36,7 +36,7 @@ def trigger_gmail_data_import_pipeline_dag(**context):
         )
 
 
-WEB_API_CONTROLLER_DAG = create_dag(
+GMAIL_CONTROLLER_DAG = create_dag(
     dag_id=DAG_ID,
     schedule_interval=os.getenv(
         GMAIL_DATA_PIPELINE_SCHEDULE_INTERVAL_ENV_NAME
@@ -44,7 +44,7 @@ WEB_API_CONTROLLER_DAG = create_dag(
 )
 
 TRIGGER_GMAIL_ETL_DAG_TASK = create_python_task(
-    WEB_API_CONTROLLER_DAG,
+    GMAIL_CONTROLLER_DAG,
     "trigger_gmail_etl_dag",
     trigger_gmail_data_import_pipeline_dag,
 )
--- a/dags/gmail_data_import_controller.py
+++ b/dags/gmail_data_import_controller.py
@@ -1,0 +1,50 @@
+# Note: DagBag.process_file skips files without "airflow" or "DAG" in them
+
+import os
+from dags.gmail_data_import_pipeline import LOGGER
+from data_pipeline.utils.pipeline_file_io import get_yaml_file_as_dict
+from data_pipeline.utils.dags.data_pipeline_dag_utils import (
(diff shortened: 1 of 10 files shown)


dags/gmail_data_import_pipeline.py
```diff
@@ -107,25 +85,25 @@ def is_end2end_test():
     )
 
 
-def get_gmail_credentials(is_e2e_test: bool) -> GmailCredentials:
+def get_gmail_credentials(is_e2e_test: bool, data_config: GmailDataConfig) -> GmailCredentials:
     if is_e2e_test:
```
de-code: It looks like this condition is now longer needed?

dags/gmail_data_import_controller.py
```diff
@@ -0,0 +1,50 @@
...
+    data_config = MultiGmailDataConfig(data_config_dict)
+    LOGGER.info("data_config: %s", data_config)
+    for gmail_data_config in data_config.gmail_data_config.values():
+        trigger_data_pipeline_dag(
+            dag_id=TARGET_DAG_ID,
+            conf=gmail_data_config
+        )
+
+
+WEB_API_CONTROLLER_DAG = create_dag(
```
de-code: best to rename

data_pipeline/gmail_data/get_gmail_data.py
```diff
@@ -99,12 +98,10 @@ def iter_link_message_thread_ids(service: Resource, user_id: str) -> Iterable[di
 
 def get_link_message_thread_ids(
         service: Resource,
-        user_id: str,
-        is_end2end_test: bool) -> pd.DataFrame:
+        user_id: str
+) -> pd.DataFrame:
     imported_timestamp = get_current_timestamp()
     messages_iterable = iter_link_message_thread_ids(service, user_id)
-    if is_end2end_test:
```
de-code, line 106: here the condition could still make sense since this was to process less data as part of the end to end test

data_pipeline/gmail_data/get_gmail_data.py
```diff
@@ -176,12 +173,10 @@ def get_dataframe_for_history_iterable(
 def get_gmail_history_details(
         service: Resource,
         user_id: str,
-        start_id: str,
-        is_end2end_test: bool) -> pd.DataFrame:
+        start_id: str
+) -> pd.DataFrame:
     imported_timestamp = get_current_timestamp()
     history_iterable = iter_gmail_history(service, user_id, start_id)
-    if is_end2end_test:
```
de-code, line 183: same here

data_pipeline/gmail_data/get_gmail_data.py
de-code: (alternatively this could also be part of the config)

HazalCiplak: As we are using the `gmail test account` we don't need it actually, we could remove it before too. I think I have added this part of code before creating test account. Count of emails in this account is less than 10, sometimes I am sending email for test and deleting them. No emails are coming to this account. What do you think should we keep it still or remove?

de-code: Okay, looks like we wouldn't benefit much from it.

de-code (approved):

PR #929, opened 2022-03-23
always process europepmc until yesterday

de-code:
part of https://github.com/elifesciences/data-hub-issues/issues/380

This allows us to save state in batches, while still processing the whole data until yesterday.


de-code:
We could reduce the max days and maybe rename it.

HazalCiplak (approved):

PR #934, opened 2022-03-24
reduced airflow min file process interval

de-code:
Perhaps 5 seconds is a good balance.
(A number out of thin air)

Let's see how long the ci job takes.


de-code:
End to end tests seem to take `10m 8s` for this ci job, vs `27m 36s` for the previous ci job.
(That is about the same as a few days ago before the change)

HazalCiplak (approved):

PR #961, opened 2022-04-23
semantic scholar pipeline

de-code:
This provides a pipeline to import data from Semantic Scholar.

As the search is limited, we are instead querying individual articles by DOI.
The DOIs will come from the EuropePMC responses.


de-code:
Some outstanding tasks:

- add support for API key
- optionally implement rate limit (the default is 100 requests / 5 minutes, with the API key it is 100 requests / 1 second)
- nice to have: implement parallel requests

With the API key, we are unlikely going to hit the limit as it is currently processed sequentially.
Without the API key we are likely going to hit the rate limit though.
 
This could be implemented separately (separate PRs).

HazalCiplak (approved):

de-code:
Raised https://github.com/elifesciences/data-hub-issues/issues/406 for the intermittent ci failures.

PR #964, opened 2022-04-28
Upgrade pylint

HazalCiplak:
https://github.com/elifesciences/data-hub-issues/issues/327


tests/dag_validation_test/crossref_pipeline_dag_validation_test.py
```diff
@@ -35,7 +35,7 @@ def test_are_tasks_around_etl_task_correctly_ordered(dagbag):
     downstream_task_ids = list(
         map(lambda task: task.task_id,
             crossref_data_etl_task.downstream_list))
-    downstream_task_ids = downstream_task_ids.sort()
+    downstream_task_ids.sort()
```
de-code, line 38: BTW `sorted` would be the functional equivalent. Not sure why we hadn't just wrapped it with that on line 35.

de-code (approved):

PR #987, opened 2022-05-18
Upgrade airflow 2.3.0

HazalCiplak:
https://github.com/elifesciences/data-hub-issues/issues/121

### DIFF
--- a/tests/end2end_test/g_spreadsheet_end_to_end_test.py
+++ b/tests/end2end_test/g_spreadsheet_end_to_end_test.py
@@ -38,19 +38,14 @@ def test_dag_runs_data_imported():
 
     AIRFLW_API.unpause_dag(TARGET_DAG_ID)
     execution_date = AIRFLW_API.trigger_dag(dag_id=DAG_ID)
-    is_running = True
-    while is_running:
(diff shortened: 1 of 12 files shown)


de-code (approved):

tests/end2end_test/g_spreadsheet_end_to_end_test.py
```diff
@@ -38,14 +38,19 @@ def test_dag_runs_data_imported():
...
-    dag_status: str
-    while True:
-        dag_status = AIRFLW_API.get_dag_status(DAG_ID, execution_date)
-        if dag_status not in {"running", "queued"}:
-            break
+    is_running = True
+    while is_running:
+        is_running = AIRFLW_API.is_dag_running(DAG_ID, execution_date)
+        if not is_running:
+            is_running = AIRFLW_API.is_triggered_dag_running(
```
de-code, line 45: I believe this section is slightly different. It is because the controller DAG is triggering sub DAGs.
Ideally we would put the loop into a function (e.g. `wait_for_dag_run_success` or similar).
Then we could call it with the controller DAG id, and then then child DAG id.

PR #995, opened 2022-05-30
include item_list in semantic scholar rec provenance

de-code:
For the evaluation it was a bit more difficult to relate the recommendation back to the events.
Including the `item_list` should make that easier.
It also provides a better information where the dois are coming from.


data_pipeline/utils/json.py
```diff
@@ -0,0 +1,26 @@
+from datetime import datetime
+
+
+def get_json_compatible_value(value):
```
de-code, line 4: partially copied and adapted from data-science-dags

data_pipeline/semantic_scholar/semantic_scholar_recommendation_pipeline.py
```diff
@@ -138,17 +146,20 @@ def get_recommendation_response_json_from_api(  # pylint: disable=too-many-argum
...
-    return get_response_json_with_provenance_from_api(
-        url,
-        params=params,
-        method='POST',
-        json_data=json_data,
-        provenance=extended_provenance,
-        session=session,
-        raise_on_status=False,
-        progress_message=progress_message
+    return get_recursive_json_compatible_value(
```
de-code, line 152: this was a "quick fix". Not sure what the right level is for calling the function, maybe it should be part of `load_given_json_list_data_from_tempdir_to_bq`.

HazalCiplak (approved):

PR #996, opened 2022-05-30
fixed webserver UI logs by setting AIRFLOW__WEBSERVER__SECRET_KEY env variable

HazalCiplak:
https://github.com/elifesciences/data-hub-issues/issues/121


docker-compose.yml
```diff
@@ -18,6 +18,9 @@ x-airflow-env:
     - AIRFLOW__API__AUTH_BACKEND=airflow.api.auth.backend.default
     - AIRFLOW__API__ENABLE_EXPERIMENTAL_API=True
     - AIRFLOW__SCHEDULER__MIN_FILE_PROCESS_INTERVAL=5
+    - AIRFLOW__WEBSERVER__SECRET_KEY='WmZHRmJwd1dCUEp6Xl4zVA=='
+    - AIRFLOW__WEBSERVER__SECRET_KEY_CMD='WmZHRmJwd1dCUEp6Xl4zVA=='
```
de-code: the `_CMD` environment variables usually replace the non `_CMD` environment counter part by providing a command to get the value instead. That is to avoid secrets being visible in environment variables. (Not an issue for development)

HazalCiplak: I am planing to define the secrets in flux cluster config if that is what you mean?

docker-compose.yml
```diff
@@ -18,6 +18,9 @@ x-airflow-env:
     - AIRFLOW__API__AUTH_BACKEND=airflow.api.auth.backend.default
     - AIRFLOW__API__ENABLE_EXPERIMENTAL_API=True
     - AIRFLOW__SCHEDULER__MIN_FILE_PROCESS_INTERVAL=5
+    - AIRFLOW__WEBSERVER__SECRET_KEY='WmZHRmJwd1dCUEp6Xl4zVA=='
```
de-code, line 21: Just adding this environment variable along seem to have worked for me.

docker-compose.yml
de-code: I could imagine that the Airflow Helm chart is already taking care of that?

de-code: But what I meant was that there should never be `AIRFLOW__WEBSERVER__SECRET_KEY` and `AIRFLOW__WEBSERVER__SECRET_KEY_CMD` at the same time. Additionally whenever there is a `_CMD` environment variable, it should be a command to get the value, not the value itself.

de-code (approved):

PR #1011, opened 2022-06-15
Related Articles pipeline

HazalCiplak:
https://github.com/elifesciences/data-hub-issues/issues/431


tests/unit_test/related_articles/related_articles_config_test.py
```diff
@@ -0,0 +1,40 @@
...
+
+GIT_REPO_URL = 'git_repo_url'
+DIRECTORY_NAME = 'directory_name'
+
+SOURCE_CONFIG = {
+    'gitRepoUrl': GIT_REPO_URL,
+    'directoryName': DIRECTORY_NAME
+}
+
+PROJECT_NAME = 'project'
```
de-code: Note: the reason why I usually add a number, like `_1`, is that the implementation could accidentally just return the config key `project` rather than the value from the dict. By using values other than the key, this would be more likely spotted. (And in t he variable name, to say it is an example and there could be others)
(Just wanted to explain my reasoning behind it)

data_pipeline/elife_article_xml/elife_article_xml_pipeline.py
```diff
@@ -0,0 +1,124 @@
...
+) -> str:
+    response_json = get_json_response_from_url(url=source_config.git_repo_url)
+    for folder in response_json['tree']:
+        if folder['path'] == source_config.directory_name:
+            return folder['url']
+
+
+def iter_xml_file_url_from_git_directory(
+    source_config: ElifeArticleXmlSourceConfig,
+    processed_file_url_list: Iterable[str]
```
de-code: I think the type would need to be `Sequence` or `Set`. (`Iterable` wouldn't work because it can be iterated only once)

de-code: Just a thought, skipping processed files could also be done in a separate function.
(Otherwise the function could maybe be named `iter_remaining_xml_file....`)

data_pipeline/elife_article_xml/elife_article_xml_pipeline.py
```diff
@@ -0,0 +1,124 @@
...
+
+def iter_decoded_xml_file_content(
+    article_xml_url_list: Iterable[str]
+) -> Iterable[str]:
+    for article_xml_url in article_xml_url_list:
+        response_json = get_json_response_from_url(url=article_xml_url)
+        if response_json['encoding'] == 'base64':
+            yield base64.b64decode(response_json['content']).decode('utf-8')
+        else:
+            LOGGER.info('File is not decoded base64, file url: %s', article_xml_url)
```
de-code: did you find examples that are not base64 encoded?

tests/unit_test/elife_article_xml/elife_article_xml_pipeline_test.py
```diff
@@ -0,0 +1,324 @@
...
+        get_json_response_from_url_mock: MagicMock
+    ):
+        get_json_response_from_url_mock.return_value = GITHUB_TREE_REPONSE_1
+        actual_return_value = get_url_of_xml_file_directory_from_repo(SOURCE_CONFIG_1)
+        assert actual_return_value == MATCHING_URL
+
+
+@pytest.mark.usefixtures("get_url_of_xml_file_directory_from_repo_mock")
+class TestIterXmlFileUrlFromGitDirectory:
+    def test_should_return_url_list_only_for_not_empty_and_non_processed_files(
```
de-code, line 253: we then also should have a test, that it is skipping zero length files, and already processed files.

tests/unit_test/elife_article_xml/elife_article_xml_pipeline_test.py
```diff
@@ -0,0 +1,324 @@
...
+        xml_string = '<article><front><article-meta></article-meta></front></article>'
+        return_value = get_article_json_data_from_xml_string_content(xml_string)
+        assert return_value == {}
+
+    def test_should_return_dict_with_related_article_if_it_exists(self):
+        # pylint: disable=line-too-long
+        xml_string = '''
+<article><front><article-meta>\
+<related-article ext-link-type="doi" id="ra1" \
+related-article-type="related_article_type_1" href="related_article_doi_1"/>\
```
de-code: we need to add the xlink namespace and handle it

tests/unit_test/elife_article_xml/elife_article_xml_pipeline_test.py
```diff
@@ -0,0 +1,324 @@
...
+        return_value = get_article_json_data_from_xml_string_content(xml_string)
+        assert return_value == {}
+
+    def test_should_return_empty_dict_if_the_xml_has_empty_article_meta(self):
+        xml_string = '<article><front><article-meta></article-meta></front></article>'
+        return_value = get_article_json_data_from_xml_string_content(xml_string)
+        assert return_value == {}
+
+    def test_should_return_dict_with_related_article_if_it_exists(self):
+        # pylint: disable=line-too-long
```
de-code: do we still need that exclusion? (should be on a line to not exclude the rule for the remaining of the file)

tests/unit_test/elife_article_xml/elife_article_xml_pipeline_test.py
```diff
@@ -0,0 +1,324 @@
...
+        xml_string = '''
+<article><front><article-meta>\
```
de-code: I don't know what the best way is of long strings in code.
I sometimes use this syntax (string will be concatenated):

```python
str_value = (
    'text1'
    'text2'
    'text3
)
```

Or (sometimes with a different delimiter):

```python
str_value = ''.join([
    'text1',
    'text2',
    'text3
])
```

tests/unit_test/utils/json_test.py
```diff
@@ -28,3 +32,103 @@ def test_should_format_datetime_within_a_list(self):
...
+    def test_should_transform_dict_value_inside_a_list(self):
+        assert get_recursively_transformed_object(
+            [{'key1': 'old value1'}],
+            key_value_transform_fn=lambda key, value: (
+                (key, 'new value1' if key == 'key1' else value)
+            )
+        ) == [{'key1': 'new value1'}]
+
+
+class TestRemoveKeyWithNullValue:
```
de-code, line 79: maybe would be good to add a test to check that:

- `None` is removed from a list
- a list item is removed, that only becomes after empty after None is removed from the item itself (e.g. `[{'key1': None}]`)

data_pipeline/elife_article_xml/elife_article_xml_pipeline.py
HazalCiplak: changed the file name to `iter_unprocessed_xml_file_url_from_git_directory`

data_pipeline/elife_article_xml/elife_article_xml_pipeline.py
HazalCiplak: no didn’t actually, do you think shouldn’t we check it?

de-code: In that case maybe it is better to fail the pipeline so that we become more aware of it?
You could change the if to an assert for example.

tests/unit_test/elife_article_xml/elife_article_xml_pipeline_test.py
HazalCiplak: testing it too

tests/unit_test/elife_article_xml/elife_article_xml_pipeline_test.py
```diff
@@ -0,0 +1,339 @@
...
+        xml_string = '<article></article>'
+        return_value = get_article_json_data_from_xml_string_content(xml_string)
+        assert return_value == {}
+
+    def test_should_return_empty_dict_if_the_xml_has_empty_article_meta(self):
+        xml_string = '<article><front><article-meta></article-meta></front></article>'
+        return_value = get_article_json_data_from_xml_string_content(xml_string)
+        assert return_value == {}
+
+    def test_should_return_dict_with_related_article_if_it_exists(self):
```
de-code: We should add a test with multiple related articles as I suspect there will be an issue with the href namespace.

data_pipeline/elife_article_xml/elife_article_xml_pipeline.py
de-code: (just leaves the type annotation)

data_pipeline/elife_article_xml/elife_article_xml_pipeline.py
```diff
@@ -0,0 +1,141 @@
...
+    xml_root = ET.fromstring(xml_string)
+    parsed_dict = parse_xml_and_return_it_as_dict(xml_root)
+    parsed_dict = get_bq_compatible_json_dict(parsed_dict)
+    if parsed_dict:
+        article_meta_dict = parsed_dict['article']['front'][0]['article_meta'][0]
+        for key in article_meta_dict.copy().keys():
+            if key != 'related_article':
+                article_meta_dict.pop(key, None)
+        if 'related_article' in article_meta_dict:
+            for key in article_meta_dict['related_article'][0].copy().keys():
```
de-code: The accessing with zero look a bit suspicious. We could probably come up with a theoretical issue for `front` and `article-meta`. Although in practise it is probably fine. But here I would think it would cause an issue. You could of course just iterate over the items, e.g.

```python
for related_article in article_meta_dict.get('related_article', []):
    ...
```

To change remove the namespace though, you could also just modify `get_bq_compatible_transformed_key_value` as that is its main purpose. Either like you have done here:

```python
if 'href' in key:
    return 'href', value
```

Or using a regular expression.

```python
key = re.sub(r'^{[^}]*}', '', key)
...
```

tests/unit_test/elife_article_xml/elife_article_xml_pipeline_test.py
```diff
@@ -0,0 +1,420 @@
...
+        ])
+        return_value = get_article_json_data_from_xml_string_content(xml_string)
+        assert return_value == {}
+
+    def test_should_return_dict_with_related_article_and_article_if_if_they_exist(self):
+        xml_string = ''.join([
+            '<article article-type="article_type_1" xmlns:xlink="http://www.w3.org/1999/xlink">',
+            '<front><article-meta>',
+            '<article-id pub-id-type="publisher-id">publisher_id_1</article-id>',
+            '<article-id pub-id-type="doi">doi_1</article-id>',
```
de-code, line 361: It will be good to also test that we remove other elements at different levels, i.e. as a child of `article` and `front`.

tests/unit_test/elife_article_xml/elife_article_xml_pipeline_test.py
```diff
@@ -0,0 +1,420 @@
...
+            '<related-article ext-link-type="doi" id="ra1" ',
+            'related-article-type="related_article_type_1" xlink:href="related_article_doi_1"/>',
+            '</article-meta></front>',
+            '</article>'
+        ])
+
+        return_value = get_article_json_data_from_xml_string_content(xml_string)
+        assert return_value == {
```
de-code, line 369: I believe the structure has changed to what we had in the test yesterday. It no longer contains the `article`, `front` and `article-meta`.
i.e. if we wanted to include other elements outside `article-meta`, then the structure would likely need to change.

data_pipeline/elife_article_xml/elife_article_xml_pipeline.py
```diff
@@ -0,0 +1,152 @@
...
+        project_name=project_name,
+        dataset_name=dataset_name,
+        table_name=table_name
+    ):
+        processed_file_url_list = get_single_column_value_list_from_bq_query(
+            project_name=project_name,
+            query=f'''
+                SELECT articles.article_xml.article_xml_url
+                FROM
+                `elife-data-pipeline.{dataset_name}.elife_article_xml_related_articles` AS articles
```
de-code: probably use the other placeholders here too, for project and table

data_pipeline/elife_article_xml/elife_article_xml_pipeline.py
```diff
@@ -0,0 +1,152 @@
...
+        )
+    else:
+        processed_file_url_list = []
+
+    article_xml_url_list = iter_unprocessed_xml_file_url_from_git_directory(
+        source_config=config.source,
+        processed_file_url_list=processed_file_url_list
+    )
+
+    for xml_file_url_and_content in iter_xml_file_url_and_decoded_content(article_xml_url_list):
```
de-code: minor: maybe just use `for xml_file_url, xml_content in ...`? (would make it a bit easier to read)

tests/unit_test/elife_article_xml/elife_article_xml_pipeline_test.py
```diff
@@ -0,0 +1,479 @@
...
+        response_mock.raise_for_status.side_effect = RuntimeError()
+        with pytest.raises(GitHubRateLimitError):
+            get_json_response_from_url(url=GIT_REPO_URL)
+
+    def test_should_raise_exceptions_from_raise_for_status(
+        self,
+        requests_mock: MagicMock
+    ):
+        response_mock = requests_mock.get.return_value
+        response_mock.raise_for_status.side_effect = RuntimeError()
```
de-code, line 236: I wonder whether is is worth using `requests.exceptions.HTTPError` (to replicate the exception that is usually thrown)

data_pipeline/elife_article_xml/elife_article_xml_pipeline.py
```diff
@@ -0,0 +1,159 @@
...
+from data_pipeline.utils.json import (
+    get_recursively_transformed_object,
+    remove_key_with_null_value
+)
+from data_pipeline.utils.xml import parse_xml_and_return_it_as_dict
+
+LOGGER = logging.getLogger(__name__)
+
+
+class GitHubRateLimitError(RuntimeError):
```
de-code: Maybe `IOError` would be an even better base class (requests' `RequestException` extends that)

data_pipeline/elife_article_xml/elife_article_xml_pipeline.py
```diff
@@ -0,0 +1,159 @@
...
+    pass
+
+
+def get_json_response_from_url(url: str) -> Any:
+    response = requests.get(url=url)
+    if response.status_code == 403:
+        raise GitHubRateLimitError()
```
de-code, line 36: maybe it is worth adding the response content as part of the message (then we would log it when we catch the exception)

de-code (approved):

PR #1015, opened 2022-06-27
Adding Git hub access token

HazalCiplak:
https://github.com/elifesciences/data-hub-issues/issues/446

### DIFF
--- a/data_pipeline/elife_article_xml/elife_article_xml_pipeline.py
+++ b/data_pipeline/elife_article_xml/elife_article_xml_pipeline.py
@@ -1,5 +1,5 @@
 import base64
-from typing import Iterable, Any, Optional, Sequence, Tuple
+from typing import Iterable, Any, Mapping, Optional, Sequence, Tuple
 import logging
 import xml.etree.ElementTree as ET
 import requests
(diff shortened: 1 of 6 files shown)


data_pipeline/elife_article_xml/elife_article_xml_pipeline.py
```diff
@@ -29,8 +29,16 @@ class GitHubRateLimitError(requests.RequestException):
...
 
 
-def get_json_response_from_url(url: str) -> Any:
-    response = requests.get(url=url)
+def get_json_response_from_url(
+    url: str,
+    headers: Mapping[str, str] = None,
+) -> Any:
+    if headers:
+        LOGGER.info("API call with access token...")
```
de-code: Maybe nitpicking, but here we just know whether headers have been configured, not whether it using an access token. If you pass in the `printable_mapping`, then we could log it. (Also together with the URL would probably be useful)

sample_data_config/elife-article-xml/elife-article-xml.config.yaml
```diff
@@ -3,6 +3,11 @@ elifeArticleXml:
     source:
       gitRepoUrl: 'https://api.github.com/repos/elifesciences/elife-article-xml/git/trees/master'
       directoryName: 'articles'
+      headers:
+        parametersFromFile:
+          # parameters will be skipped if referenced file does not exist
+          - parameterName: x-api-key
```
de-code: I believe the parameter name is `Authorization`. And the value is not just the API token itself, but `token <TOKEN>`.
i.e. that would need to be reflected in the secret content, and its name.

Perhaps that is one disadvantage of the more generic "headers".

data_pipeline/elife_article_xml/elife_article_xml_pipeline.py
```diff
@@ -29,8 +29,16 @@ class GitHubRateLimitError(requests.RequestException):
...
+        response = requests.get(url=url, headers=headers)
```
de-code: I believe you can always call it this way, even if `headers` is `None`.
(i.e. you don't need the condition)

sample_data_config/elife-article-xml/elife-article-xml.config.yaml
HazalCiplak: Yes I was looking for this, just find out and read you comment :D

de-code (approved):

PR #1022, opened 2022-07-07
Add article type and path to article xml pipeline

HazalCiplak:
https://github.com/elifesciences/data-hub-issues/issues/454

### DIFF
--- a/data_pipeline/elife_article_xml/elife_article_xml_pipeline.py
+++ b/data_pipeline/elife_article_xml/elife_article_xml_pipeline.py
@@ -56,7 +56,7 @@ def get_url_of_xml_file_directory_from_repo(
 def iter_unprocessed_xml_file_url_and_path_from_git_directory(
     source_config: ElifeArticleXmlSourceConfig,
     processed_file_url_list: Sequence[str]
-) -> Iterable[str]:
+) -> Iterable[Tuple[str, str]]:
     response_json = get_json_response_from_url(
         url=get_url_of_xml_file_directory_from_repo(source_config=source_config),
         headers=source_config.headers.mapping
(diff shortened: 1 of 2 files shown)


data_pipeline/elife_article_xml/elife_article_xml_pipeline.py
```diff
@@ -53,7 +53,7 @@ def get_url_of_xml_file_directory_from_repo(
             return folder['url']
 
 
-def iter_unprocessed_xml_file_url_from_git_directory(
+def iter_unprocessed_xml_file_url_and_path_from_git_directory(
     source_config: ElifeArticleXmlSourceConfig,
     processed_file_url_list: Sequence[str]
 ) -> Iterable[str]:
```
de-code: the return type annotation is probably no longer accurate

data_pipeline/elife_article_xml/elife_article_xml_pipeline.py
```diff
@@ -64,20 +64,24 @@ def iter_unprocessed_xml_file_url_from_git_directory(
...
             if article_xml_url['url'] not in processed_file_url_list:
-                yield article_xml_url['url']
+                yield (article_xml_url['url'], article_xml_url['path'])
 
 
-def iter_xml_file_url_and_decoded_content(
+def iter_xml_file_url_path_and_decoded_content(
     source_config: ElifeArticleXmlSourceConfig,
-    article_xml_url_list: Iterable[str]
+    article_xml_url_and_path_list: Iterable[str]
```
de-code: type annotation here too

data_pipeline/elife_article_xml/elife_article_xml_pipeline.py
```diff
@@ -64,20 +64,24 @@ def iter_unprocessed_xml_file_url_from_git_directory(
...
 ) -> Iterable[Tuple]:
```
de-code: I just realised that the type annotation could be expanded, e.g. `Iterable[Tuple[str, str, str]]`.

With three fields a named tuple could also be useful, not sure if necessary here though.

data_pipeline/elife_article_xml/elife_article_xml_pipeline.py
```diff
@@ -116,7 +120,10 @@ def get_article_json_data_from_xml_string_content(
                         if key not in ('related_article', 'article_id'):
                             article_meta_dict.pop(key, None)
                     LOGGER.info(article_meta_dict)
-                return article_meta_dict
+                return {
+                    'article_type': parsed_dict['article']['article_type'],
+                    **article_meta_dict
```
de-code: I would usually put the expansion of the original dict first, that would make it clear that `article_type` is always set from this place. Otherwise if `article_meta_dict` contained an `article_type` key, it would be used instead. (It is probably impossible for this pipeline, but may not be obvious just reading the code)

de-code (approved):
would fix the type annotations before merging

PR #1042, opened 2022-08-04
Get campaigns via Twitter Ads API

HazalCiplak:
https://github.com/elifesciences/data-hub-issues/issues/475

### DIFF
--- a/data_pipeline/twitter_ads_api/twitter_ads_api_pipeline.py
+++ b/data_pipeline/twitter_ads_api/twitter_ads_api_pipeline.py
@@ -6,17 +6,21 @@ from data_pipeline.twitter_ads_api.twitter_ads_api_config import (
     TwitterAdsApiConfig,
     TwitterAdsApiSourceConfig
 )
-from data_pipeline.utils.data_store.bq_data_service import load_given_json_list_data_from_tempdir_to_bq
+from data_pipeline.utils.data_store.bq_data_service import (
+    load_given_json_list_data_from_tempdir_to_bq
+)
 from data_pipeline.utils.json import remove_key_with_null_value
 
 
 def get_client_from_twitter_ads_api(
     source_config: TwitterAdsApiSourceConfig
 ) -> Client:
-    return Client(consumer_key=source_config.secrets.mapping['api_key'],
-                consumer_secret=source_config.secrets.mapping['api_secret'],
-                access_token=source_config.secrets.mapping['access_token'],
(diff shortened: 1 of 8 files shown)


sample_data_config/twitter-ads-api/twitter-ads-api.config.yaml
```diff
@@ -0,0 +1,19 @@
+twitterAdsApi:
+  - dataPipelineId: twitter_ads_api_campaign_elife_journal
+    source:
+      resource: '/11/accounts/18ce53up4ko/campaigns'
+      secrets:
+        parametersFromFile:
+          - parameterName: api_key
+            filePathEnvName: TWITTER_API_KEY_FILE_PATH
```
de-code, line 8: would include "ADS" in the env name, as you have for the config file.

HazalCiplak: This api keys are from twitter api not just for Ads api. We can use it for twitter api too if it is needed.

HazalCiplak: There is not specific key creation for Twitter Ads API.

de-code: Okay, I didn't realise.

data_pipeline/twitter_ads_api/twitter_ads_api_pipeline.py
```diff
@@ -0,0 +1,49 @@
+from typing import Any
```
de-code, line 1: should we have tests for this module?

de-code (approved):

data_pipeline/twitter_ads_api/twitter_ads_api_pipeline.py
HazalCiplak: I did not find an avantage to add test for the current functions? I am planing to add a test in next PR as there will be different kind of functions to test which would make more sense then.

de-code: I guess it would be that we passed around the values correctly.
Perhaps an end to end test? (I guess we would need to add the credentials to vault)

HazalCiplak: yes can do when Scott is back to create one.

PR #1050, opened 2022-08-08
Add headers support to semantic scholar recommendations

de-code:
part of https://github.com/elifesciences/data-hub-issues/issues/481

### DIFF
--- a/data_pipeline/utils/pipeline_utils.py
+++ b/data_pipeline/utils/pipeline_utils.py
@@ -102,7 +102,10 @@ def get_response_json_with_provenance_from_api(  # noqa pylint: disable=too-many
         if progress_message
         else ''
     )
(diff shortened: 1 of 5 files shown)


data_pipeline/utils/pipeline_utils.py
```diff
@@ -102,7 +114,14 @@ def get_response_json_with_provenance_from_api(  # noqa pylint: disable=too-many
         if progress_message
         else ''
     )
-    LOGGER.info('requesting url%s: %r %r (%r)', progress_message_str, method, url, params)
+    printable_headers = get_default_printable_mapping_with_secrets(
```
de-code, line 117: I thought it was worth adding this default behaviour to avoid the log message saying "headers=None" when actually headers were passed in.

HazalCiplak (approved):

PR #1051, opened 2022-08-09
Twitter Ads API audience reach per campaign

HazalCiplak:
https://github.com/elifesciences/data-hub-issues/issues/477

### DIFF
--- a/data_pipeline/twitter_ads_api/twitter_ads_api_config.py
+++ b/data_pipeline/twitter_ads_api/twitter_ads_api_config.py
@@ -1,5 +1,5 @@
 
-from typing import Iterable, NamedTuple
+from typing import Iterable, NamedTuple, Sequence
 
 from data_pipeline.utils.pipeline_config import (
     BigQuerySourceConfig,
@@ -42,8 +42,9 @@ class TwitterAdsApiConfig(NamedTuple):
         )
 
     @staticmethod
-    def from_dict(config_dict: dict) -> 'TwitterAdsApiConfig':
-        item_config_list = config_dict['twitterAdsApi']
-        return TwitterAdsApiConfig._from_item_dict(
-            item_config_list[0]
-        )
+    def parse_config_list_from_dict(config_dict: dict) -> Sequence['TwitterAdsApiConfig']:
+        item_config_dict_list = config_dict['twitterAdsApi']
+        return [
+            TwitterAdsApiConfig._from_item_dict(item_config_dict)
+            for item_config_dict in item_config_dict_list
+        ]
--- a/data_pipeline/twitter_ads_api/twitter_ads_api_config.py
+++ b/data_pipeline/twitter_ads_api/twitter_ads_api_config.py
@@ -1,4 +1,4 @@
-from typing import NamedTuple, Sequence
+from typing import NamedTuple, Optional, Sequence
 
 from data_pipeline.utils.pipeline_config import (
     BigQuerySourceConfig,
@@ -9,7 +9,7 @@ from data_pipeline.utils.pipeline_config import (
 
 class TwitterAdsApiParameterValuesConfig(NamedTuple):
     from_bigquery: BigQuerySourceConfig = {}
-    start_time_value: str = ''
+    start_time_value: Optional[str] = None
 
     @staticmethod
     def from_dict(parameter_values_config_dict: dict) -> 'TwitterAdsApiParameterValuesConfig':
@@ -22,9 +22,9 @@ class TwitterAdsApiParameterValuesConfig(NamedTuple):
 
 
 class TwitterAdsApiParameterNamesForConfig(NamedTuple):
-    bigquery_value: str = ''
-    start_time: str = ''
-    end_time: str = ''
+    bigquery_value: Optional[str] = None
+    start_time: Optional[str] = None
+    end_time: Optional[str] = None
 
     @staticmethod
     def from_dict(parameter_names_for_config_dict: dict) -> 'TwitterAdsApiParameterNamesForConfig':
--- a/data_pipeline/twitter_ads_api/twitter_ads_api_config.py
+++ b/data_pipeline/twitter_ads_api/twitter_ads_api_config.py
@@ -11,18 +11,22 @@ from data_pipeline.utils.pipeline_config import (
 class TwitterAdsApiSourceConfig(NamedTuple):
     resource: str
     secrets: MappingConfig = MappingConfig.from_dict({})
-    param_from_bigquery: BigQuerySourceConfig = {}
-    param_names: Iterable[str] = []
+    param_value_from_bigquery: BigQuerySourceConfig = {}
+    param_name_for_bq_value: str = ''
+    param_name_for_start_time: str = ''
+    param_name_for_end_time: str = ''
 
     @staticmethod
     def from_dict(source_config_dict: dict) -> 'TwitterAdsApiSourceConfig':
         return TwitterAdsApiSourceConfig(
             resource=source_config_dict['resource'],
             secrets=MappingConfig.from_dict(source_config_dict['secrets']),
-            param_from_bigquery=BigQuerySourceConfig.from_dict(
-                source_config_dict.get('paramFromBigQuery', {})
+            param_value_from_bigquery=BigQuerySourceConfig.from_dict(
+                source_config_dict.get('paramValueFromBigQuery', {})
             ),
-            param_names=source_config_dict.get('paramNames', []),
+            param_name_for_bq_value=source_config_dict.get('paramNameForBqValue', ''),
+            param_name_for_start_time=source_config_dict.get('paramNameForStartTime', ''),
+            param_name_for_end_time=source_config_dict.get('paramNameForEndTime', '')
         )
 
 
--- a/data_pipeline/twitter_ads_api/twitter_ads_api_config.py
+++ b/data_pipeline/twitter_ads_api/twitter_ads_api_config.py
@@ -11,7 +11,7 @@ from data_pipeline.utils.pipeline_config import (
 class TwitterAdsApiSourceConfig(NamedTuple):
     resource: str
     secrets: MappingConfig = MappingConfig.from_dict({})
-    bigquery: BigQuerySourceConfig = {}
+    param_from_bigquery: BigQuerySourceConfig = {}
     param_names: Iterable[str] = []
 
     @staticmethod
@@ -19,8 +19,10 @@ class TwitterAdsApiSourceConfig(NamedTuple):
         return TwitterAdsApiSourceConfig(
             resource=source_config_dict['resource'],
             secrets=MappingConfig.from_dict(source_config_dict['secrets']),
+            param_from_bigquery=BigQuerySourceConfig.from_dict(
+                source_config_dict.get('paramFromBigQuery', {})
+            ),
             param_names=source_config_dict.get('paramNames', []),
-            bigquery=BigQuerySourceConfig.from_dict(source_config_dict.get('bigQuery', {}))
         )
 
 
--- a/data_pipeline/twitter_ads_api/twitter_ads_api_config.py
+++ b/data_pipeline/twitter_ads_api/twitter_ads_api_config.py
@@ -58,7 +58,7 @@ class TwitterAdsApiSourceConfig(NamedTuple):
 
     @staticmethod
     def from_dict(source_config_dict: dict) -> 'TwitterAdsApiSourceConfig':
-        if source_config_dict.get('apiQueryParameters', {}):
+        if source_config_dict.get('apiQueryParameters'):
             return TwitterAdsApiSourceConfig(
                 resource=source_config_dict['resource'],
                 secrets=MappingConfig.from_dict(source_config_dict['secrets']),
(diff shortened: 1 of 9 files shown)


data_pipeline/twitter_ads_api/twitter_ads_api_pipeline.py
```diff
@@ -24,35 +30,101 @@ def get_client_from_twitter_ads_api(
...
+def get_param_dict_from_required_params_and_value_from_bq(
+    source_config: TwitterAdsApiSourceConfig,
+    value_from_bq: str,
+    start_time: str,
+    end_time: str
+):
     return {
-        **remove_key_with_null_value(response.body),
-        'provenance': provenance
+        source_config.required_params['paramNameForBqValue']: value_from_bq,
```
de-code: That the key is `paramNameForBqValue` feels like config and could maybe be moved to the config parsing?

data_pipeline/utils/pipeline_config.py
```diff
@@ -25,15 +25,15 @@ class PipelineEnvironmentVariables:
 
 
 class BigQuerySourceConfig(NamedTuple):
-    project_name: str
-    sql_query: str
+    project_name: str = ''
```
de-code: if string values are optional then it would be better to declare them as `Optional` and set to `None` rather than an empty string.
Although in this case it would be better to make the usage of `BigQuerySourceConfig` optional.
The advantage is that we then either have an instance with all of the required fields in `BigQuerySourceConfig` or we don't. It would be caught at config parsing time whether something is missing.

de-code: The more we can validate at config parsing time the better.

data_pipeline/utils/pipeline_utils.py
```diff
@@ -24,19 +24,22 @@ def fetch_single_column_value_list_for_bigquery_source_config(
...
-        )
-    except google.cloud.exceptions.NotFound:
-        if bigquery_source_config.ignore_not_found:
-            LOGGER.info('caught not found, returning empty list')
-            return []
-        raise
-    LOGGER.debug('value_list: %r', value_list)
-    LOGGER.info('length of value_list: %r', len(value_list))
-    return value_list
+    if bigquery_source_config.project_name and bigquery_source_config.sql_query:
```
de-code: Based on the https://github.com/elifesciences/data-hub-core-airflow-dags/pull/1051/files#r943236494 I would suggest to not change this function.

If at all then we could make `bigquery_source_config` itself optional, e.g.:

```python
if not bigquery_source_config:
    return []
...
```

But then it seems that condition should at the caller maybe?

You already have a condition in `iter_bq_compatible_json_response_from_resource_with_provenance` based on `required_params`. I guess in that case, configuring the query should be required?

tests/unit_test/twitter_ads_api/twitter_ads_api_config_test.py
```diff
@@ -2,66 +2,152 @@
...
+                {
+                    **ITEM_CONFIG_DICT_2,
+                    'source': {
+                        **SOURCE_CONFIG_2,
+                        'secrets': secrets
+                    }
+                }
+            )
+        )
+        assert config[0].source.secrets.mapping['key1'] == secrets['key1']
```
de-code: this test looks identical to the test above, except the assert on a specific key?

data_pipeline/twitter_ads_api/twitter_ads_api_pipeline.py
```diff
@@ -24,35 +30,101 @@ def get_client_from_twitter_ads_api(
...
+    if source_config.required_params:
+        value_list_from_bq = fetch_single_column_value_list_for_bigquery_source_config(
+            source_config.param_value_from_bigquery
+        )
+        today = date.today()
+        yesterday = today - timedelta(days=1)
+        for value_from_bq in value_list_from_bq:
+            params_dict = get_param_dict_from_required_params_and_value_from_bq(
+                source_config=source_config,
+                value_from_bq=value_from_bq,
```
de-code, line 93: I believe we currently do not have a unit test checking that we are passing in the correct values to `get_param_dict_from_required_params_and_value_from_bq_mock`?

data_pipeline/twitter_ads_api/twitter_ads_api_pipeline.py
```diff
@@ -24,35 +30,101 @@ def get_client_from_twitter_ads_api(
...
+                # earliest campaign creation date for initial load:
+                start_time='2018-05-01',
```
de-code: we can certainly hard code values. But it might still be worth putting that into config as that is where I would probably look for it. What do you think?

data_pipeline/twitter_ads_api/twitter_ads_api_pipeline.py
```diff
@@ -24,35 +30,101 @@ def get_client_from_twitter_ads_api(
...
+        source_config.required_params['paramNameForStartTime']: start_time,
+        source_config.required_params['paramNameForEndTime']: end_time
     }
 
 
+def iter_bq_compatible_json_response_from_resource_with_provenance(
+    source_config: TwitterAdsApiSourceConfig
+) -> Any:
+    if source_config.required_params:
+        value_list_from_bq = fetch_single_column_value_list_for_bigquery_source_config(
```
de-code, line 87: it might be worth moving the contents of this if statement to a separate function? (wouldn't necessarily have to affect the test, which would still test this function)

sample_data_config/twitter-ads-api/twitter-ads-api.config.yaml
```diff
@@ -1,7 +1,36 @@
...
+      resource: '/11/stats/accounts/18ce53up4ko/reach/campaigns'
+      paramValueFromBigQuery:
+        projectName: 'elife-data-pipeline'
+        sqlQuery: |-
+          SELECT DISTINCT t_data.id
+          FROM `elife-data-pipeline.hc_dev.v_latest_twitter_campaign_details_elife_journal`
+          JOIN UNNEST(data) AS t_data
+          WHERE t_data.id IN ('git3k', 'gwdrg')
+      requiredParams:
+        paramNameForBqValue: campaign_ids
```
de-code: minor / comment:
In general I would maybe try to avoid abbreviations. Sometimes it may not be clear for someone unfamiliar (although I guess in this case it is). The other argument would be consistency. So far all other references to "BigQuery" in the config seem to be `BigQuery` not `Bq`.

`parameter` is mostly spelled out, except for `params` (that may have been a mistake). I guess we used `params` because that is the name of the parameter within the requests library. Maybe we should change it? Perhaps even to make clear that these are query parameters passed in via the URL.

Here, what do you think of changing the name of the container to something like `requiredParameterNameFor`, then the fields could be `bigQueryValue`, `startTime` and `endTime`. Maybe we could also drop the `required` from the name? Maybe `apiQueryParameterNameFor` (not sure if that is too long or is any clearer)

sample_data_config/twitter-ads-api/twitter-ads-api.config.yaml
```diff
@@ -1,7 +1,36 @@
...
+  #   target:
+  #     projectName: 'elife-data-pipeline'
+  #     datasetName: '{ENV}'
+  #     tableName: twitter_campaign_details_elife_journal
+
+  - dataPipelineId: twitter_ads_api_campaign_reach_elife_journal
     source:
-      resource: '/11/accounts/18ce53up4ko/campaigns'
+      resource: '/11/stats/accounts/18ce53up4ko/reach/campaigns'
+      paramValueFromBigQuery:
```
de-code: Just wondering whether this should be moved out of "source" or whether "source" should have a separate api category. Just to make it more clear what affects the API and what not. For example the resource, parameters and secrets are for the API.
Whereas the BigQuery data is separate from the API. It isn't the main source of the pipeline, it rather informs it what to query. A bit like a state file might.
What do you think?

HazalCiplak: structure has been changed as we talked

data_pipeline/twitter_ads_api/twitter_ads_api_config.py
```diff
@@ -7,21 +7,63 @@
...
+        return TwitterAdsApiParameterValuesConfig(
+            from_bigquery=BigQuerySourceConfig.from_dict(
+                parameter_values_config_dict.get('fromBigQuery', {})
+            ),
+            start_time_value=parameter_values_config_dict.get('startTimeValue', '')
+        )
+
+
+class TwitterAdsApiParameterNamesForConfig(NamedTuple):
+    bigquery_value: str = ''
```
de-code: Similar to a previous comment, I would generally mark them as `Optional[str]` and default to `None`. I believe a blank string is usually not a good default value.

In this iteration you could even argue that they are all mandatory?

data_pipeline/twitter_ads_api/twitter_ads_api_config.py
```diff
@@ -7,21 +7,63 @@
 )
 
 
+class TwitterAdsApiParameterValuesConfig(NamedTuple):
+    from_bigquery: BigQuerySourceConfig = {}
+    start_time_value: str = ''
```
de-code: Same comment regarding default value for the `start_time_value`

tests/unit_test/twitter_ads_api/twitter_ads_api_config_test.py
```diff
@@ -2,66 +2,176 @@
...
+        )
+
+    def test_should_read_from_bigquery_sql_query_and_project_name_if_defined(self):
+        config = TwitterAdsApiConfig.parse_config_list_from_dict(
+            get_config_for_item_config_dict(
+                ITEM_CONFIG_DICT_1,
+                ITEM_CONFIG_DICT_WITH_API_QUERY_PARAMETERS
+            )
+        )
+        assert config[1].source.api_query_parameters.parameter_values.from_bigquery.sql_query == (
```
de-code: I find this test a bit more difficult to follow because we have two items and need to associate index 1 with the one with query parameters and index 0 without.

This would probably be clearer as two separate tests?

tests/unit_test/twitter_ads_api/twitter_ads_api_config_test.py
```diff
@@ -2,66 +2,176 @@
...
+            START_TIME_VALUE_1
+        )
+        assert config[0].source.api_query_parameters.parameter_values.start_time_value == (
+            ''
+        )
+
+    def test_should_read_from_bigquery_sql_query_and_project_name_if_defined(self):
+        config = TwitterAdsApiConfig.parse_config_list_from_dict(
+            get_config_for_item_config_dict(
+                ITEM_CONFIG_DICT_1,
```
de-code: maybe we could rename that config constant to say it doesn't contain API query parameters?

tests/unit_test/twitter_ads_api/twitter_ads_api_config_test.py
```diff
@@ -2,66 +2,176 @@
...
+        )
+        assert config[1].source.api_query_parameters.parameter_values.from_bigquery.sql_query == (
+            SQL_QUERY
+        )
+        assert (
+            config[1].source.api_query_parameters.parameter_values.from_bigquery.project_name == (
+                PROJECT_NAME
+            )
+        )
+        assert config[0].source.api_query_parameters.parameter_values.from_bigquery.sql_query == (
```
de-code, line 140: In the case where we haven't defined `api_query_parameters`, I thought the idea was that `api_query_parameters` would then be `None`.
Then one can check `if source_counfig.source.api_query_parameters`.
And if it is is not None, then all required fields within it will be populated.

data_pipeline/twitter_ads_api/twitter_ads_api_pipeline.py
HazalCiplak: I am not sure why we should do that? In if statement only thing we are doing is calling functions.

tests/unit_test/twitter_ads_api/twitter_ads_api_config_test.py
HazalCiplak: You are right, but I think I prefer the check exact value in the fields too.

data_pipeline/twitter_ads_api/twitter_ads_api_pipeline.py
HazalCiplak: Lets keep this in mind to discuss it on Monday.

HazalCiplak:
I am planing to merge this to develop to be able to continue to next task. We can make necessary changes in next PR as it will be related to this.

sample_data_config/twitter-ads-api/twitter-ads-api.config.yaml
```diff
@@ -1,7 +1,25 @@
 twitterAdsApi:
-  - dataPipelineId: twitter_ads_api_campaign_elife_journal
+  # - dataPipelineId: twitter_ads_api_campaign_elife_journal
```
de-code, line 2: should we enable this part to test that this is still working?

tests/unit_test/twitter_ads_api/twitter_ads_api_config_test.py
de-code: I meant this comment relating to the configuration of `api_query_parameter` to be None, which I think you addressed. Checking individual values is fine.

sample_data_config/twitter-ads-api/twitter-ads-api.config.yaml
HazalCiplak: It is working I have test it on my local and it is on staging too and working.

PR #1065, opened 2022-08-22
Implementation of start_date and end_period usage 

HazalCiplak:
https://github.com/elifesciences/data-hub-issues/issues/484

### DIFF
--- a/data_pipeline/twitter_ads_api/twitter_ads_api_config.py
+++ b/data_pipeline/twitter_ads_api/twitter_ads_api_config.py
@@ -10,7 +10,6 @@ from data_pipeline.utils.pipeline_config import (
 
 class TwitterAdsApiParameterValuesConfig(NamedTuple):
     from_bigquery: BigQuerySourceConfig = {}
-    api_min_start_date: Optional[str] = '2015-01-02'
     max_period_in_days: Optional[int] = 0
     placement_value: Optional[Sequence[str]] = []
     period_batch_size_in_days: Optional[int] = 0
@@ -21,7 +20,6 @@ class TwitterAdsApiParameterValuesConfig(NamedTuple):
             from_bigquery=BigQuerySourceConfig.from_dict(
                 parameter_values_config_dict.get('fromBigQuery', {})
             ),
-            api_min_start_date=parameter_values_config_dict.get('apiMinStartDate', '2015-01-02'),
             max_period_in_days=parameter_values_config_dict.get('maxPeriodInDays', 0),
             period_batch_size_in_days=parameter_values_config_dict.get('periodBatchSizeInDays', 0),
             placement_value=parameter_values_config_dict.get('placementValue', [])
--- a/data_pipeline/twitter_ads_api/twitter_ads_api_config.py
+++ b/data_pipeline/twitter_ads_api/twitter_ads_api_config.py
@@ -1,4 +1,3 @@
-from datetime import date
 from typing import NamedTuple, Optional, Sequence
 
 from data_pipeline.utils.pipeline_config import (
--- a/data_pipeline/twitter_ads_api/twitter_ads_api_config.py
+++ b/data_pipeline/twitter_ads_api/twitter_ads_api_config.py
@@ -1,3 +1,4 @@
+from datetime import date
 from typing import NamedTuple, Optional, Sequence
 
 from data_pipeline.utils.pipeline_config import (
@@ -9,6 +10,7 @@ from data_pipeline.utils.pipeline_config import (
 
 class TwitterAdsApiParameterValuesConfig(NamedTuple):
     from_bigquery: BigQuerySourceConfig = {}
+    api_min_start_date: Optional[str] = '2015-01-02'
     max_period_in_days: Optional[int] = 0
     placement_value: Optional[Sequence[str]] = []
     period_batch_size_in_days: Optional[int] = 0
@@ -19,6 +21,7 @@ class TwitterAdsApiParameterValuesConfig(NamedTuple):
             from_bigquery=BigQuerySourceConfig.from_dict(
                 parameter_values_config_dict.get('fromBigQuery', {})
             ),
+            api_min_start_date=parameter_values_config_dict.get('apiMinStartDate', '2015-01-02'),
             max_period_in_days=parameter_values_config_dict.get('maxPeriodInDays', 0),
             period_batch_size_in_days=parameter_values_config_dict.get('periodBatchSizeInDays', 0),
             placement_value=parameter_values_config_dict.get('placementValue', [])
--- a/data_pipeline/twitter_ads_api/twitter_ads_api_config.py
+++ b/data_pipeline/twitter_ads_api/twitter_ads_api_config.py
@@ -11,6 +11,7 @@ class TwitterAdsApiParameterValuesConfig(NamedTuple):
     from_bigquery: BigQuerySourceConfig = {}
     start_date_value: Optional[str] = None
     end_date_value: Optional[str] = None
+    end_period_per_days: Optional[int] = None
     placement_value: Optional[Sequence[str]] = []
 
     @staticmethod
@@ -21,6 +22,7 @@ class TwitterAdsApiParameterValuesConfig(NamedTuple):
             ),
             start_date_value=parameter_values_config_dict.get('startDateValue', None),
             end_date_value=parameter_values_config_dict.get('endDateValue', None),
+            end_period_per_days=parameter_values_config_dict.get('endPeriodPerDays', None),
             placement_value=parameter_values_config_dict.get('placementValue', [])
         )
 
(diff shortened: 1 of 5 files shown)


data_pipeline/twitter_ads_api/twitter_ads_api_config.py
```diff
@@ -9,8 +9,7 @@
 
 class TwitterAdsApiParameterValuesConfig(NamedTuple):
     from_bigquery: BigQuerySourceConfig = {}
-    start_date_value: Optional[str] = None
-    end_date_value: Optional[str] = None
+    ending_period_per_day: Optional[int] = 0
```
de-code: naming:
  The suffix `in_days` might be better? (`per` suggests to me a breakdown or denominator)
  At the beginning `max` might be better? (a period has a start and an end, `ending period` may suggest to me there are multiple period and this refers to the the last period within a larger context, e.g. "ending period of 20th century")

data_pipeline/twitter_ads_api/twitter_ads_api_pipeline.py
```diff
@@ -85,36 +84,87 @@ def get_param_dict_from_api_query_parameters(
     return result
 
 
+def get_final_end_date(
+    api_query_parameters_config: TwitterAdsApiApiQueryParametersConfig,
+    initial_start_date_value: datetime
+) -> datetime:
+    yesterday_date = get_yesterdays_date()
+    end_date_by_period = initial_start_date_value + timedelta(
```
de-code: minor naming: `end_date_of_period`? Or `period_end_date` (I often aim to put the type at the end, although sometimes that doesn't read well)

tests/unit_test/twitter_ads_api/twitter_ads_api_pipeline_test.py
```diff
@@ -294,6 +292,58 @@ def test_should_return_response_json_with_provenance(
...
+    def test_should_return_final_end_date_as_yesterday_if_the_ending_period_days_is_in_future(
+        self,
+        get_yesterdays_date_mock: MagicMock
+    ):
+        api_query_parameters = API_QUERY_PARAMETERS_WITH_SINGLE_PLACEMENT_VALUE._replace(
+            parameter_values=PARAMETER_VALUES_WITH_PLACEMENT._replace(
+                ending_period_per_day=10  # in 10 days
+            )
+        )
+        get_yesterdays_date_mock.return_value = datetime(2022, 8, 5).date()  # in 4 days
```
de-code: just a comment: you could also use the `date` class directly

tests/unit_test/twitter_ads_api/twitter_ads_api_pipeline_test.py
```diff
@@ -294,6 +292,58 @@ def test_should_return_response_json_with_provenance(
...
+            api_query_parameters_config=api_query_parameters,
+            initial_start_date_value=datetime.strptime('2022-08-01', '%Y-%m-%d').date()
+        )
+        assert actual_return_value == datetime.strptime('2022-08-11', '%Y-%m-%d').date()
+
+
+class TestGetEndDateValueOfPeriod:
+    def test_should_return_period_end_date_as_end_date(self):
+        actual_return_value = get_end_date_value_of_period(
+            start_date_value=datetime.strptime('2022-08-01', '%Y-%m-%d').date(),
```
de-code: you could also use `date.fromisoformat`

data_pipeline/twitter_ads_api/twitter_ads_api_pipeline.py
```diff
@@ -85,36 +84,87 @@ def get_param_dict_from_api_query_parameters(
     return result
 
 
+def get_final_end_date(
```
de-code: To me it wasn't clear what `final` means here and what the difference between `get_final_end_date` and `get_end_date_value_of_period` was. I suppose in part it is also because `period` may refer to any period.
It might help if you could name the two periods.
I don't know what good names would be. Names that come to mind but may not be good: `overall period` and `batch period`.

data_pipeline/twitter_ads_api/twitter_ads_api_pipeline.py
```diff
@@ -85,36 +84,87 @@ def get_param_dict_from_api_query_parameters(
...
+    return final_end_date_value
+
+
+def get_end_date_value_of_period(
+    start_date_value: datetime,
+    final_end_date_value: datetime,
+    days_in_period: int
+) -> datetime:
+    period_end_date_value = start_date_value + timedelta(days=days_in_period)
+    if period_end_date_value > final_end_date_value:
```
de-code: maybe this if / else block could be reused somehow. If those were numbers, then you could use `min`. So maybe `get_min_date(date1: date, date2: date) -> date`?

Or including adding the days:

`add_days_with_max_date(start_date: date, days: int, max_date: date) -> date`?

data_pipeline/twitter_ads_api/twitter_ads_api_pipeline.py
```diff
@@ -85,36 +84,87 @@ def get_param_dict_from_api_query_parameters(
...
+    )
+    if end_date_by_period > yesterday_date:
+        final_end_date_value = yesterday_date
+    else:
+        final_end_date_value = end_date_by_period
+    return final_end_date_value
+
+
+def get_end_date_value_of_period(
+    start_date_value: datetime,
```
de-code: I guess the type should be `date`?
I would maybe just call it `start_date`.

data_pipeline/twitter_ads_api/twitter_ads_api_pipeline.py
```diff
@@ -85,36 +84,87 @@ def get_param_dict_from_api_query_parameters(
...
+            api_query_parameters_config.parameter_values.from_bigquery
+        ))
+        LOGGER.debug("dict_value_list_from_bq: %r", dict_value_list_from_bq)
+        for dict_value_from_bq in dict_value_list_from_bq:
+            entity_id_from_bq = dict_value_from_bq['entity_id']
+            start_date_str_from_bq = dict_value_from_bq['start_date']
+            initial_start_date_value = (
+                datetime.strptime(start_date_str_from_bq, '%Y-%m-%d').date()
+            )
+            if initial_start_date_value <= datetime.strptime('2015-01-01', '%Y-%m-%d').date():
```
de-code: Is the hardcoded date intentional?

data_pipeline/twitter_ads_api/twitter_ads_api_pipeline.py
```diff
@@ -85,36 +84,87 @@ def get_param_dict_from_api_query_parameters(
...
+                    entity_id_from_bq
+                )
+                continue
+            final_end_date_value = get_final_end_date(
+                api_query_parameters_config=api_query_parameters_config,
+                initial_start_date_value=initial_start_date_value
+            )
+
+            placement_value_list = api_query_parameters_config.parameter_values.placement_value
+            if not api_query_parameters_config.parameter_names_for.placement:
```
de-code, line 140: My reading of this is that the assumption is that if we do not have a placement value, then we do not need to break up the period. It might be useful to add an assert that `ending_period_per_day` (or the final name) was not specified.

de-code: Actually thinking about it again, it is not `ending_period_per_day`. It is the `7 days` that isn't configured.

data_pipeline/twitter_ads_api/twitter_ads_api_pipeline.py
HazalCiplak: `min_api_start_date`

data_pipeline/twitter_ads_api/twitter_ads_api_pipeline.py
HazalCiplak: `period_batch_size_in_days`

data_pipeline/twitter_ads_api/twitter_ads_api_pipeline.py
HazalCiplak: added filter to SQL query instead of using this variable as we talked in the meeting as an option

data_pipeline/twitter_ads_api/twitter_ads_api_config.py
HazalCiplak: `maxPeriodInDays`

data_pipeline/twitter_ads_api/twitter_ads_api_pipeline.py
HazalCiplak: `period_max_end_date`

data_pipeline/twitter_ads_api/twitter_ads_api_pipeline.py
HazalCiplak: `get_current_final_end_date`

PR #1085, opened 2022-09-20
[Generic web api] Hypothesis pipeline failure

HazalCiplak:
https://github.com/elifesciences/data-hub-issues/issues/498

### DIFF
--- a/tests/unit_test/generic_web_api/generic_web_api_data_etl_test.py
+++ b/tests/unit_test/generic_web_api/generic_web_api_data_etl_test.py
@@ -102,6 +102,17 @@ class TestGetItemList:
         )
         assert actual_response == [data]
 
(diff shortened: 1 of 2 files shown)


tests/unit_test/generic_web_api/generic_web_api_data_etl_test.py
```diff
@@ -102,6 +102,17 @@ def test_should_return_all_data_when_no_data_path_key(
         )
         assert actual_response == [data]
 
+    def test_should_return_key_list_even_the_list_is_empty(
```
de-code, line 154: I guess ideally we'd have a test at the level of `generic_web_api_data_etl` because that is where the problem could have materialised.

de-code (approved):

PR #1147, opened 2022-12-06
added surveymonkey id list manually

HazalCiplak:
https://github.com/elifesciences/data-hub-issues/issues/535


de-code (approved):
A bit surprised that we didn't need to update a test :-)

PR #1153, opened 2022-12-19
use same s3 bucket with deployment env path

de-code:
see also #1152 

this allows to use a deployment env like `de_dev` without requiring a separate bucket

(only include done sample config, it could be applied to all of them)

### DIFF
--- a/sample_data_config/crossref-event/crossref-event-data-pipeline.config.yaml
+++ b/sample_data_config/crossref-event/crossref-event-data-pipeline.config.yaml
@@ -13,7 +13,7 @@ stateFile:
 schemaFile:
   url:
   bucket: 'ci-elife-data-pipeline'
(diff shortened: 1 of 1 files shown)


HazalCiplak (approved):

de-code:
I suggest to maybe before merging:

- ask Scott to create a new S3 bucket, e.g. `elife-data-pipeline`
- update config to use the new bucket
- update all of the sample config

(Then we would avoid mixed content in the bucket)

HazalCiplak:
I think, it will be good to have a prefix on the bucket name, like `dev-elife-data-pipeline` or `test-elife-data-pipeline`. What do you think?

de-code:
> I think, it will be good to have a prefix on the bucket name, like `dev-elife-data-pipeline` or `test-elife-data-pipeline`. What do you think?

I actually thought we could then also move prod and staging over in the future too. (Then we could delete the other three buckets)

PR #1166, opened 2023-01-05
Web API: ignore invalid timestamp value /w schema

de-code:
related to https://github.com/elifesciences/data-hub-issues/issues/577

### DIFF
--- a/data_pipeline/generic_web_api/transform_data.py
+++ b/data_pipeline/generic_web_api/transform_data.py
@@ -88,17 +88,23 @@ def standardize_record_keys(record_object):
         return new_list
 
 
+def _is_valid_timestamp_string(timestamp_str: str) -> bool:
+    try:
+        if dateparser.parse(timestamp_str):
+            return True
+    except BaseException:
+        pass
+    return False
+
+
 def _get_valid_timestamp_string_or_none(
     timestamp_str: str,
     key: str
 ) -> Optional[str]:
     if timestamp_str is None:
         return None
-    try:
-        if dateparser.parse(timestamp_str):
-            return timestamp_str
-    except BaseException:
-        pass
+    if _is_valid_timestamp_string(timestamp_str):
+        return timestamp_str
     LOGGER.warning('ignoring invalid timestamp value (key: %r): %r', key, timestamp_str)
     return None
 
--- a/data_pipeline/generic_web_api/transform_data.py
+++ b/data_pipeline/generic_web_api/transform_data.py
@@ -88,7 +88,10 @@ def standardize_record_keys(record_object):
         return new_list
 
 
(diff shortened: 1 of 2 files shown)


tests/unit_test/generic_web_api/transform_data_test.py
```diff
@@ -0,0 +1,44 @@
...
+                    "name": "ts",
+                    "type": "TIMESTAMP"
+                }
+            ]
+        )
+        assert result == {'ts': None}
+
+    def test_should_set_timestamp_to_none_if_invalid_value(self):
+        result = filter_record_by_schema(
+            record_object={'ts': '0000-01-01T00:00:00Z'},
```
de-code: that is the value I actually saw being logged as invalid.

data_pipeline/generic_web_api/transform_data.py
```diff
@@ -84,6 +88,27 @@ def standardize_record_keys(record_object):
...
+def _is_valid_timestamp_string(timestamp_str: str) -> bool:
+    try:
+        if dateparser.parse(timestamp_str):
+            return True
+    except BaseException:
+        pass
+    return False
+
+
+def _get_valid_timestamp_string_or_none(
```
HazalCiplak, line 100: Just wondered why do we need to have `_` beginning of this function ? And maybe in previous (`_is_valid_timestamp_string`) one too?

tests/unit_test/generic_web_api/transform_data_test.py
HazalCiplak: It is a proper invalid date :)

tests/unit_test/generic_web_api/transform_data_test.py
```diff
@@ -0,0 +1,57 @@
...
+                    "mode": "NULLABLE",
+                    "name": "ts",
+                    "type": "TIMESTAMP"
+                }
+            ]
+        )
+        assert result == {'ts': None}
+
+    def test_should_set_timestamp_to_none_if_invalid_type(self):
+        result = filter_record_by_schema(
```
HazalCiplak, line 37: this test look same as `test_should_set_timestamp_to_none_if_invalid_format` or I couldn't see the difference

data_pipeline/generic_web_api/transform_data.py
de-code: We discussed this in our data catch-up. There are no strict rules. But my thinking was those methods, within a utility module, are there to make another function simpler. But I wouldn't want it to be used elsewhere. (Especially with the questionable use of `dateparser` which is a bit too lax)

de-code:
I made some small improvements. Could you please re-review?

HazalCiplak (approved):

PR #1198, opened 2023-02-10
Web API: Added support for record timestamp in nested list

de-code:
For the Twiiter API, it is possible to include extra data from linked ids.
In that case the response structure looks like:

```json
{
  "data": [...]
  "includes": {},
  "meta": {}
}
```

Whereas `data` is the list of Tweet data that contains for example an `author_id`.
Whereas `includes.users` could include the user summary of those author ids.

In that case it's easier to just dump the whole response to BigQuery (like we have now done for the EuropePMC response).
But then there is a single record.
We can then configure `itemTimestampKeyFromItemRoot` to be `["data", "created_at"]`, which previously would have failed as it is trying to parse a list of timestamp strings. Now it would be treated the same as if we extracted individual records from `data`.

### DIFF
--- a/data_pipeline/generic_web_api/transform_data.py
+++ b/data_pipeline/generic_web_api/transform_data.py
@@ -4,6 +4,7 @@ from typing import Any, Iterable, List, Optional, Sequence
 import dateparser
 
 from data_pipeline.generic_web_api.module_constants import ModuleConstant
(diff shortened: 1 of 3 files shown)


de-code:
I hope it is okay to merge, I am aiming to start collecting Twitter data which doesn't provide historic data beyond a week.

PR #1206, opened 2023-02-28
Toggl Pipeline issue 

HazalCiplak:
https://github.com/elifesciences/data-hub-issues/issues/572

### DIFF
--- a/data_pipeline/generic_web_api/generic_web_api_data_etl.py
+++ b/data_pipeline/generic_web_api/generic_web_api_data_etl.py
@@ -5,7 +5,7 @@ from tempfile import TemporaryDirectory
 from pathlib import Path
 import json
 from json.decoder import JSONDecodeError
-from typing import Any, Optional
+from typing import Any, Optional, Tuple, cast
 
 from botocore.exceptions import ClientError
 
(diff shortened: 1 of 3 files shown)


HazalCiplak:
I have dissabled [attr-defined] mypy error code, because it was complaining about a third party library (`requests`).

de-code:
> I have dissabled [attr-defined] mypy error code, because it was complaining about a third party library (`requests`).

Maybe in that case just ignoring the line via `# type: ignore` makes sense? (then we wouldn't ignore the linting for that for the whole project)

HazalCiplak:
> > I have dissabled [attr-defined] mypy error code, because it was complaining about a third party library (`requests`).
> 
> Maybe in that case just ignoring the line via `# type: ignore` makes sense? (then we wouldn't ignore the linting for that for the whole project)

I have added like that but there was 4 different  places for same library, I thought it is not ignoring the linting just mypy as added in mypy.ini file. If it effects pylint, I can update but I am not sure how it does

de-code:
> > > I have dissabled [attr-defined] mypy error code, because it was complaining about a third party library (`requests`).
> > 
> > 
> > Maybe in that case just ignoring the line via `# type: ignore` makes sense? (then we wouldn't ignore the linting for that for the whole project)
> 
> I have added like that but there was 4 different places for same library, I thought it is not ignoring the linting just mypy as added in mypy.ini file. If it effects pylint, I can update but I am not sure how it does

Maybe we can chat about. It wouldn't affect pylint. But mypy linting for other cases. `4` doesn't seem that much. But I don't completely understand where and what is causing it.

PR #1251, opened 2023-07-25
Get VOR publication date from articles xml repo

HazalCiplak:
https://github.com/elifesciences/data-hub-issues/issues/738


de-code (approved):
Personally I think would prefer a single target table for the parsed XML files.

For me the reasons:
- Avoids inefficient fetching and parsing the same XML files twice
- Reduces maintenance overhead, only one pipeline that can fail and one target table to monitor

The downside (or benefit of the separate config):
- One time effort to manage transition

Maybe one option could be to make the second config more generic including all of the fields and deprecate the first config (allowing time to update the views).

HazalCiplak:
> Personally I think would prefer a single target table for the parsed XML files.

I have tried to parse and write whole XML into BQ but I got the error `google.api_core.exceptions.BadRequest: 400 PATCH Fields must contain the allowed characters, and be at most 300 characters long.` And didn’t want to spend time for finding them and replacing with the allowed characters.

de-code:
> > Personally I think would prefer a single target table for the parsed XML files.
> 
> I have tried to parse and write whole XML into BQ but I got the error `google.api_core.exceptions.BadRequest: 400 PATCH Fields must contain the allowed characters, and be at most 300 characters long.` And didn’t want to spend time for finding them and replacing with the allowed characters.

Yes, that is fine. But we could have a longer list of selected fields to still have a single target table.

sample_data_config/elife-article-xml/elife-article-xml.config.yaml
```diff
@@ -8,7 +8,24 @@ elifeArticleXml:
           # parameters will be skipped if referenced file does not exist
           - parameterName: Authorization
             filePathEnvName: GITHUB_API_AUTHORIZATION_FILE_PATH
+      # XML elements that we are interested in (by replaceing '-' with '_')
+      searchedXmlElements: ['related_article', 'article_id', 'article_categories']
```
de-code: Actually I just realise, `selected` might be a better terms. And maybe `Element` is slightly confusing because it happens after conversion which you documented. But maybe it would be better represented in the name? e.g. `selectedNormalizedXmlElements` or similar?

de-code:
Was the latest commit necessary or was it just to rerun ci?

HazalCiplak:
> Was the latest commit necessary or was it just to rerun ci?

I have tried to rerun ci  manually couple of times but it didn’t work, so for testing it I have made that commit :)

PR #1270, opened 2023-08-15
Added BigQuery to OpenSearch data pipeline

de-code:


de-code:
It is failing to update settings even though settings haven't changed. Need to fix that.

Also need to implement state handling.

PR #1277, opened 2023-08-24
OpenSearch: Added timeout configuration

de-code:
part of https://github.com/elifesciences/data-hub-issues/issues/763

It seemed valuable since we sometimes even received timeouts with a batch size of 100.
This also increases the default timeout to 60 seconds


de-code:
The Generic Web API seem to be taking a long time and then timing out. I had re-run it a few times.

de-code:
Actually it seems to be the people API that is causing issues.
It is currently unavailable.

PR #1312, opened 2023-10-18
Updated sample config to import from EuropePMC separately

de-code:
part of https://github.com/elifesciences/data-hub-issues/issues/794

I thought it is still useful to update the example config too.
I also added `10.1101/2022.09.28.509871` which isn't available in S2.

Some potential future improvement:

- rename `fieldNamesFor` to `keyPathsFor`
- introduce `KeyPath` class
- move index settings out of pipeline config (e.g. to top level)

### DIFF
--- a/sample_data_config/opensearch/bigquery-to-opensearch.yaml
+++ b/sample_data_config/opensearch/bigquery-to-opensearch.yaml
@@ -1,5 +1,5 @@
 bigQueryToOpenSearch:
-  - dataPipelineId: bigquery_to_opensearch_data_pipeline_s2_1
+  - dataPipelineId: bigquery_to_opensearch_data_pipeline_s2
     source:
       bigQuery:
         projectName: 'elife-data-pipeline'
@@ -108,7 +108,7 @@ bigQueryToOpenSearch:
                     full_text_list:
                       type: "nested"
     batchSize: 100
-  - dataPipelineId: bigquery_to_opensearch_data_pipeline_europepmc_2
+  - dataPipelineId: bigquery_to_opensearch_data_pipeline_europepmc
     source:
       bigQuery:
         projectName: 'elife-data-pipeline'
@@ -205,7 +205,7 @@ bigQueryToOpenSearch:
         operationMode: 'update'
         upsert: True
     batchSize: 100
-  - dataPipelineId: bigquery_to_opensearch_data_pipeline_sciety_3
+  - dataPipelineId: bigquery_to_opensearch_data_pipeline_sciety
     source:
       bigQuery:
         projectName: 'elife-data-pipeline'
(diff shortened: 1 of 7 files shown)


HazalCiplak (approved):

de-code:
I didn't realise about the [nested](https://opensearch.org/docs/latest/field-types/supported-field-types/nested/) type, which isn't so obvious.

Also it seems better to use separate objects for each source (s2, europepmc, sciety).
So I am planning to change the structure and populate a new index (although it will then take a long time to populate S2 and EuropePMC again).

Any thoughts?

sample_data_config/opensearch/bigquery-to-opensearch.yaml
```diff
@@ -74,33 +81,145 @@ bigQueryToOpenSearch:
...
+                      method:
+                        # Note: we need Lucene for filter support
+                        name: "hnsw"
+                        space_type: "l2"
+                        engine: "lucene"
+                        parameters:
+                          # Default values: https://opensearch.org/docs/latest/search-plugins/knn/knn-index/
+                          ef_construction: 512
+                          m: 16
+                europepmc:
```
de-code, line 96: It would probably be more clear to configure the index settings outside the pipeline. We could also update the mappings again when we come to EuropePMC but that doesn't seem ideal either.

sample_data_config/opensearch/bigquery-to-opensearch.yaml
```diff
@@ -74,33 +81,145 @@ bigQueryToOpenSearch:
...
+                  ARRAY(
+                      SELECT AS STRUCT
+                          version.source,
+                          version.id,
+                          version.versionNumber AS version_number,
+                          version.firstPublishDate AS first_publication_date,
+                          ARRAY(
+                              SELECT publication_type
+                              FROM UNNEST(pubTypeList.pubType) AS publication_type
+                          ) AS publication_type_list
```
de-code: Add _string_

sample_data_config/opensearch/bigquery-to-opensearch.yaml
```diff
@@ -74,33 +81,145 @@ bigQueryToOpenSearch:
...
+                          author.firstName AS first_name,
+                          author.fullName AS full_name,
+                          author.initials,
+                          author.lastName AS last_name,
+                          ARRAY(
+                              SELECT affiliation
+                              FROM UNNEST(author.authorAffiliationDetailsList.authorAffiliation) AS affiliation
+                          ) AS affiliation_string_list
+                      FROM UNNEST(europepmc_response.authorList.author) AS author
+                  ) AS authors,
```
de-code: Should be author_list

PR #1321, opened 2023-11-02
Web API: Retrieve Specter v1 Embeddings via title and abstract from Semantic Scholar API

de-code:
part of https://github.com/elifesciences/data-hub-issues/issues/798


de-code:
So far drafted how the configuration might look like.
(Since the Web API is used in the end to end test, this is in a separate config file)

HazalCiplak (approved):

de-code:
Managed to load embeddings this way. More work to be done to make this scale and limit number of papers sent to API (max 16 per request).

PR #1328, opened 2023-11-09
Web API: Use include/exclude for BigQuery Source

de-code:
part of https://github.com/elifesciences/data-hub-issues/issues/798

This uses include/exclude for the recently introduced BigQuery Source config of the Web API


sample_data_config/web-api/draft-web-api-s2-embeddings.config.yaml
```diff
@@ -5,18 +5,27 @@ webApi:
...
+              '10.1101/2022.09.28.509871'
+            )
+      exclude:
+        bigQuery:
+          ignoreNotFound: true
+          projectName: 'elife-data-pipeline'
+          sqlQuery:
+            SELECT response.paper_id
+            FROM `elife-data-pipeline.{ENV}.test_s2_specter_v1_embeddings` AS response
+        keyFieldName: 'paper_id'
```
de-code: We talked about the `keyFieldName` being within `exclude`. I explained that by being required when `exclude` is configured. However, you pointed out that it relates to the field name produced by the `include` query. So maybe could do with a better name or more thinking.

data_pipeline/utils/pipeline_config.py
```diff
@@ -26,40 +38,111 @@ class PipelineEnvironmentVariables:
...
+    def from_dict(source_config_dict: BigQuerySourceConfigDict) -> 'BigQuerySourceConfig':
         return BigQuerySourceConfig(
             project_name=source_config_dict['projectName'],
             sql_query=source_config_dict['sqlQuery'],
             ignore_not_found=source_config_dict.get('ignoreNotFound', False)
         )
 
 
-class BigQueryTargetConfig(NamedTuple):
+@dataclass(frozen=True)
```
de-code, line 56: You raised why we are converting from `NamedTuple` to `dataclass`.
My reasoning was that some things, like extending base classes, can't be done with `NamedTuple`.
That is why I am tempted to convert all usage to `dataclass`.

de-code:
I have used the opportunity to add more typing of the configuration dictionaries. Perhaps that could have been a separate PR.

data_pipeline/generic_web_api/generic_web_api_config_typing.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+class WebApiDataUrlConfigDict(TypedDict):
+    urlExcludingConfigurableParameters: str
+    configurableParameters: NotRequired[WebApiConfigurableParametersConfigDict]
+    parametersFromEnv: NotRequired[Sequence[ParameterFromEnvConfigDict]]  # Note: not used anymore
+    parametersFromFile: NotRequired[Sequence[ParameterFromFileConfigDict]]
+
+
+class SchemaFileConfigDict(TypedDict):
```
de-code, line 39: The `SchemaFileConfigDict` and `StateFileConfigDict` could maybe be merged.

PR #1336, opened 2023-11-16
Web API: rename URL builder to request builder to reflect its function

de-code:
This renames "[Dynamic] URL builder" to "Dynamic Request Builder" and related parameters.

That also includes a configuration change (that we paired on).

Old:

```yaml
    urlSourceType:
      name: 'civi'
      sourceTypeSpecificValues:
        fieldsToReturn:
          - external_identifier
          - tag
```

New:

```yaml
    requestBuilder:
      name: 'civi'
      parameters:
        fieldsToReturn:
          - external_identifier
          - tag
```

It is still backwards compatible to allow for a smoother config change transition.


de-code:
Still not everything renamed. With some names I am not quite sure.
Would be good to hear what you think.

HazalCiplak:
Overall, the changes look good to me. I'm curious, why did you find the `request` keyword better than `url`? Because I think until the request is made, it's not a request; it's just a URL. I'm not sure, maybe it's my English.

de-code:
> Overall, the changes look good to me. I'm curious, why did you find the `request` keyword better than `url`? Because I think until the request is made, it's not a request; it's just a URL. I'm not sure, maybe it's my English.

It is true, that it is isn't the actual HTTP request.
The main motivation for renaming it is that it no longer is just the URL. It can now also handle the body, and also includes information about the method.
So I am thinking of it building what will contribute to the request overall.

For example we are using the `session.request` method with the following docstring:

>  Constructs a :class:`Request <Request>`, prepares it and sends it.

And starts with:

```python
        req = Request(
            method=method.upper(),
            url=url,
            headers=headers,
            files=files,
            data=data or {},
            json=json,
            params=params or {},
            auth=auth,
            cookies=cookies,
            hooks=hooks,
        )
```

At that time it isn't sent yet either. But it is close to it.

So I guess I am seeing what we are doing as something similar. Only that it is focused on the dynamic part, rather than the static configuration (e.g. `headers` are in the config, but `url` and `json` are something we build).

HazalCiplak:
I see, yes that makes sense. Thank you.

de-code:
This should now have all of the related changes.

tests/unit_test/generic_web_api/generic_web_api_config_test.py
```diff
@@ -110,3 +111,39 @@ def test_should_read_batch_size(self):
...
+            CiviWebApiDynamicRequestBuilder
+        )
+        assert web_api_config.dynamic_request_builder.request_builder_parameters == {
+            'fieldsToReturn': ['field1', 'field2']
+        }
+
+    def test_should_read_deprecated_request_builder_name_and_parameters(self):
+        web_api_config = WebApiConfig.from_dict({
+            **MINIMAL_WEB_API_CONFIG_DICT,
+            'urlSourceType': {
```
HazalCiplak, line 136: Didn’t we update these fields? In the previous test it looks updated. Does that mean we are keeping the both version?

de-code: Yes, that is the test for backwards compatibility. We can remove that after the change is deployed and we have updated the config.

PR #1338, opened 2023-11-17
Create separate Airflow DAGs from config items

de-code:
part of https://github.com/elifesciences/data-hub-issues/issues/810

There are some challenges:

- How do we ensure the DAG list is updated when the configuration changes?
- Are DAGs deleted when the config item is removed? (And should they be deleted?)
- Each DAG should also use the up-to-date config when triggered

This is how it works:

- Config is read to generate DAGs, passing in the `dataPipelineId` as a key
- Each DAG will read the config, looking for the `dataPipelineId`:
  - That ensures that the DAG is using the up-to-date config (even if changed since DAG creation)
  - Deleted config wouldn't delete the DAG, but the DAG run would fail, not causing undesired effects (DAG can be deleted manually)


de-code:
This is an initial draft demonstrating separate Web API configs as their own DAGs.
(They do not currently work though)

de-code:
Maybe doing that for the CSV pipeline would be more useful now, we could configure different schedules and maybe reduce the memory issue that way. Although the CSV pipelines do run quite frequently.

image

image

de-code:
This seems to be mostly working now (still could do with some refactory).
Although in ci the end to end test was failing, with `OSError: Dag id Web_API.observer_published_research_article not found`
It is passing for me locally.

PR #1401, opened 2024-02-14
Web API: Ingest from multiple Crossref prefixes

de-code:
part of https://github.com/elifesciences/data-hub-issues/issues/854

This is to allow ingesting from multiple Crossref prefixes, with the same Web API pipeline (rather than copying it).

It is using the existing BigQuery source values to configure the list of DOI prefixes.

### DIFF
--- a/data_pipeline/generic_web_api/generic_web_api_data_etl.py
+++ b/data_pipeline/generic_web_api/generic_web_api_data_etl.py
@@ -443,19 +443,39 @@ def process_web_api_data_etl_batch_with_batch_source_value(
         LOGGER.info('completed batch: %d', 1 + batch_index)
 
 
+def should_iterate_over_source_values(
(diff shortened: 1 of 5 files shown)


de-code:
Next we will make sure, that we keep the placeholder values when preparing the next request parameters (for pagination)

de-code:
Confirmed that the state file from manually running it is there as expected:

```console
$ gsutil ls s3://ci-elife-data-pipeline/airflow-config/generic-web-api/ci-crossref-metadata-state/
s3://ci-elife-data-pipeline/airflow-config/generic-web-api/ci-crossref-metadata-state/10.31235.json
$ gsutil cat s3://ci-elife-data-pipeline/airflow-config/generic-web-api/ci-crossref-metadata-state/10.31235.json
2024-02-16 10:10:12+0000
````

PR #1429, opened 2024-04-18
Reduce CSV pipeline memory usage (with fixes)

HazalCiplak:
https://github.com/elifesciences/data-hub-issues/issues/904

Reduces CSV pipeline memory usage.
Fixes issues previously introduced by #1421


de-code:
When starting to download the file first, we started to see an error that the file was already closed.
We fixed that by using a context manager.
Maybe that was the actual underlying error with boto in this case.

PR #1444, opened 2024-05-15
Web API: Temporarily use people api for end to end tests

de-code:
This is due to the observer API being unavailable at the moment.

(It seemed to me that this might take a bit longer to resolve)

So I moved the people API up to be used for end to end tests instead.


de-code:
I hope it is okay to merge it for now.

PR #1488, opened 2024-07-29
Use KubernetesPodOperator with local Kubernetes environment for development

de-code:
part of https://github.com/elifesciences/data-hub-issues/issues/961

Remaining issues / tasks:

- [ ] Automatic Kubernetes config without having to update the IP manually
- [ ] Workflow using local Docker image


de-code:
This currently worked locally, after manually editing the `k3s-config/kubeconfig.yaml` file.
Changing the IP in `server: https://172.19.0.8:6443` to the IP found via the `docker network inspect` command...

```console
# get network ids
$ docker network ls
NETWORK ID     NAME                                 DRIVER    SCOPE
6840f3bbe3f1   data-hub-core-airflow-dags_default   bridge    local
...

$ docker network inspect 6840f3bbe3f1
...
# search for `data-hub-core-airflow-dags-k3s-server-1`
```

We could also use `kubectl` locally by passing in the config:

```console
$ kubectl --kubeconfig=k3s-config/kubeconfig.yaml get nodes
...

$ kubectl --kubeconfig=k3s-config/kubeconfig.yaml get pods -A
...
```

We may not see logs, if the Airflow task fails too quickly. We can look at the logs on the file system in that case via `make ci-end2end-test-logs` (and `make ci-end2end-test-clear-logs`).

de-code:
It didn't work with a local image as it was trying to pull the image rather than use the one we have already built locally.

de-code:
For some reasons, after adding the volume `/var/run/docker.sock` from host to container, there is no sign of that within the container.

```console
$ docker compose exec k3s-server sh -c 'ls -la /var/run/'
...
total 4
drwxr-xr-x 3 root root   60 Jul 29 14:55 .
drwxr-xr-x 1 root root 4096 Jul 29 10:54 ..
drwxr-xr-x 2 root root   80 Jul 29 14:56 netns
```
Currently, `/run` and `/var/run` are marked as `tmpfs` but the above still holds true even after removing that.

dags/kubernetes_pipeline.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def create_kubernetes_pipeline_dags() -> Sequence[airflow.DAG]:
+    dags = []
+    with create_dag(
+        dag_id='kubernetes_pipeline',
+        dagrun_timeout=timedelta(days=1),
+        tags=['Kubernetes']
+    ) as dag:
+        image = 'elifesciences/data-hub-core-dags-dev'
```
de-code, line 24: I thought you reverted this, maybe not pushed the change?

PR #1492, opened 2024-08-07
OpenSearch ingest pipeline tests

de-code:
Related to https://github.com/elifesciences/data-hub-issues/issues/963

This makes it easier to ensure that the ingest pipelines work correctly before updating the ingest pipeline.

### DIFF
--- a/data_pipeline/opensearch/bigquery_to_opensearch_pipeline.py
+++ b/data_pipeline/opensearch/bigquery_to_opensearch_pipeline.py
@@ -2,7 +2,7 @@ import dataclasses
 from datetime import datetime
 import json
 import logging
-from typing import Any, Iterable, Sequence
+from typing import Any, Iterable, Sequence, Tuple
 
 from opensearchpy import OpenSearch
 import opensearchpy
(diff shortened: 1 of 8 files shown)


de-code:
This should be complete now. It will log failing tests with some info.

Only problem is that it is failing quickly and Airflow has the issue of not showing logs in that case. Perhaps only locally.

data_pipeline/opensearch/bigquery_to_opensearch_pipeline.py
```diff
@@ -78,10 +80,102 @@ def get_opensearch_client(opensearch_target_config: OpenSearchTargetConfig) -> O
     )
 
 
+def get_opensearch_ingest_pipeline_tests_body(
```
HazalCiplak: I was thinking should we have a separated module for these functions as they are for testing and not effecting directly the opensearch pipeline. Maybe undersame forlder but specifically for these functions?

de-code: The only thing with naming the the module something like `opensearch_ingest_pipeline_testing` is that the test module would then be called `opensearch_ingest_pipeline_testing_test`. What do you think?

Alternatively we could have a module for ingest pipeline related code, e.g. `opensearch_ingest_pipeline` or just `ingest_pipeline`?

de-code: I opted for `ingest_pipeline` for now but could revisit and further enhance.

HazalCiplak (approved):

PR #1509, opened 2024-08-21
Update task_id with dag_id and deployment env var

HazalCiplak:
part of https://github.com/elifesciences/data-hub-issues/issues/961


de-code (approved):

de-code:
We saw that the pod name changed and the `KubernetesPodOperator` is also doing some cleanup to the name (e.g. it converts two hyphens to one).

PR #1526, opened 2024-09-10
Add image pull policy to kubernetes pipeline config

soggy-mushroom:
part of https://github.com/elifesciences/data-hub-issues/issues/966

Currently we are using the `latest` tag. That is why we want to configure the image pull policy to `Always` so that it will always check for updates rather than just use the previously downloaded version.

We have already added the configuration, but didn't pass it to the `KubernetesPodOperator`. This PR fixes.

(In the future we want to avoid using `latest` though)


de-code (approved):

de-code:
Merging as it passed tests

PR #1542, opened 2024-10-15
Kubernetes Pipeline: config cli helper

de-code:
This is in preparation to reduce duplication.

It should show the config that we end up with.
(It could also be used to check what config is currently deployed - although limited to the Kubernetes Config)


de-code:
Now that deployment is much faster there seem to be less use for this helper

PR #1548, opened 2024-10-23
Web API: make retry configurable

de-code:
part of https://github.com/elifesciences/data-hub-issues/issues/995

We've seen too many request errors and it would be good to retry with some delay.


data_pipeline/generic_web_api/generic_web_api_config_typing.py
```diff
@@ -64,6 +65,9 @@ class WebApiRecordTimestampResponseConfigDict(TypedDict):
     itemTimestampKeyFromItemRoot: NotRequired[Sequence[str]]
 
 
+OnSameNextCursorOption = Literal['Error', 'Stop', 'Continue']
```
de-code: Or should we rename to `OnSameNextCursorConfig`?

PR #1574, opened 2024-12-19
Web API: Added keywords_from_manuscript_abstract sample config

de-code:
part of https://github.com/elifesciences/data-hub-issues/issues/1046


de-code (commented):
This is extracting the keywords for the selected abstract.

Currently there is no sight of the manuscript_id or version_id in the row.
Need to think about how to tell the pipeline to add them to the provenance for example.

An alternative would be to use the abstract / text as the key and we would need to join using the abstract when we want to retrieve the keywords. But that doesn't seem ideal.

PR #1576, opened 2024-12-19
Web API: Fix `spacy_keyword_extraction` request builder not mapped

de-code:
part of https://github.com/elifesciences/data-hub-issues/issues/1046


de-code (commented):
Unfortunately we didn't spot that. Probably because the abstracts didn't include special characters and we didn't process more than one record.

PR #1596, opened 2025-01-20
Kubernetes Pipeline default config

de-code:
part of https://github.com/elifesciences/data-hub-issues/issues/1003

### DIFF
--- a/data_pipeline/kubernetes/kubernetes_pipeline_config.py
+++ b/data_pipeline/kubernetes/kubernetes_pipeline_config.py
@@ -82,7 +82,7 @@ class KubernetesPipelineConfig:  # pylint: disable=too-many-instance-attributes
                     config_dict,
                     k8s_models.V1Volume
                 )
-                for config_dict in pipeline_config_dict['volumes']
+                for config_dict in pipeline_config_dict.get('volumes', [])
             ],
             env=[
                 convert_dict_to_kubernetes_client_object(
--- a/data_pipeline/kubernetes/kubernetes_pipeline_config.py
+++ b/data_pipeline/kubernetes/kubernetes_pipeline_config.py
@@ -29,7 +29,7 @@ def convert_dict_to_kubernetes_client_object(
 def get_config_dict_for_config_dict_list(
     config_dict_list: Sequence[MappingT],
     unique_key: str
-) -> Mapping[tuple, MappingT]:
+) -> Mapping[str, MappingT]:
     return {
         config_dict[unique_key]: config_dict
         for config_dict in config_dict_list
--- a/data_pipeline/kubernetes/kubernetes_pipeline_config.py
+++ b/data_pipeline/kubernetes/kubernetes_pipeline_config.py
@@ -25,21 +25,23 @@ def convert_dict_to_kubernetes_client_object(
 
 
 def get_config_dict_for_config_dict_list(
-    config_dict_list: Sequence[dict]
-) -> Mapping[str, dict]:
+    config_dict_list: Sequence[dict],
+    unique_keys: List[str]
+) -> Mapping[tuple, dict]:
     return {
-        volume_config_dict['name']: volume_config_dict
-        for volume_config_dict in config_dict_list
+        tuple(config_dict[key] for key in unique_keys): config_dict
+        for config_dict in config_dict_list
     }
 
 
 def get_merged_config_dict_list(
     config: Optional[Sequence[dict]],
-    default_config: Optional[Sequence[dict]]
+    default_config: Optional[Sequence[dict]],
+    unique_keys: List[str]
 ) -> List[dict]:
     return list({
-        **get_config_dict_for_config_dict_list(default_config or []),
-        **get_config_dict_for_config_dict_list(config or [])
+        **get_config_dict_for_config_dict_list(default_config or [], unique_keys),
+        **get_config_dict_for_config_dict_list(config or [], unique_keys)
     }.values())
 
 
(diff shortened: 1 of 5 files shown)


data_pipeline/kubernetes/kubernetes_pipeline_config.py
```diff
@@ -23,6 +24,27 @@ def convert_dict_to_kubernetes_client_object(
...
+        env_config_dict['name']: env_config_dict
+        for env_config_dict in env_config_dict_list
+    }
+
+
+def get_merged_env_config_dict_list(
+    env: Optional[Sequence[KubernetesEnvConfigDict]],
+    default_env: Optional[Sequence[KubernetesEnvConfigDict]]
+) -> List[KubernetesEnvConfigDict]:
+    if env and default_env:
```
de-code: We could avoid this `if` statement by just adding `or []` when calling `get_env_config_dict_for_env_config_dict_list`

de-code:
Regarding volume mounts, it seems that `mountPath` should indeed be unique. Although I couldn't quite see that in the [official documentation](https://kubernetes.io/docs/concepts/storage/volumes/).

But there are some [blog posts](https://technekey.com/kubernetes-mount-multiple-volumes-into-the-same-directory/) that indicate this. There should be an error message like this one if you tried to use the same `mountPath`:

> The Pod "my-pod" is invalid: spec.containers[0].volumeMounts[1].mountPath: Invalid value: "/var/secret-data": must be unique

data_pipeline/kubernetes/kubernetes_pipeline_config.py
```diff
@@ -36,41 +60,75 @@ class KubernetesPipelineConfig:  # pylint: disable=too-many-instance-attributes
...
             volume_mounts=[
                 convert_dict_to_kubernetes_client_object(
                     config_dict,
                     k8s_models.V1VolumeMount
                 )
-                for config_dict in pipeline_config_dict['volumeMounts']
+                for config_dict in get_merged_config_dict_list(
+                    config=pipeline_config_dict.get('volumeMounts', []),
+                    default_config=default_config_dict.get('volumeMounts', []),
+                    unique_keys=['name', 'mountPath']
```
de-code: I believe the key is `mountPath` only. Because you can't mount more than one thing to the same path.

data_pipeline/kubernetes/kubernetes_pipeline_config.py
```diff
@@ -1,16 +1,19 @@
...
 
 from data_pipeline.kubernetes.kubernetes_pipeline_config_typing import (
+    KubernetesDefaultConfigDict,
     KubernetesEnvConfigDict,
     KubernetesPipelineConfigDict,
     MultiKubernetesPipelineConfigDict
 )
 from data_pipeline.utils.pipeline_config import AirflowConfig
 
+T = TypeVar("T", bound=Mapping[str, Any])
```
de-code: good use of generics.

minor: would suggest to rename to `MappingT`.

data_pipeline/kubernetes/kubernetes_pipeline_config.py
```diff
@@ -23,6 +26,27 @@ def convert_dict_to_kubernetes_client_object(
     )
 
 
+def get_config_dict_for_config_dict_list(
+    config_dict_list: Sequence[T],
+    unique_keys: List[str]
+) -> Mapping[tuple, T]:
+    return {
+        tuple(config_dict[key] for key in unique_keys): config_dict
```
de-code: I guess this would be equivalent to:

```python
tuple(unique_keys): config_dict
```

(at the moment a single key seems to be enough though)

de-code: Ignore me, we need to get the values of course.

data_pipeline/kubernetes/kubernetes_pipeline_config.py
```diff
@@ -23,6 +26,27 @@ def convert_dict_to_kubernetes_client_object(
     )
 
 
+def get_config_dict_for_config_dict_list(
+    config_dict_list: Sequence[T],
+    unique_keys: List[str]
```
de-code: Would use `Sequence` (if we still want to continue supporting multiple keys)

data_pipeline/kubernetes/kubernetes_pipeline_config_typing.py
```diff
@@ -29,6 +29,9 @@ class KubernetesPipelineConfigDict(TypedDict):
 
 class KubernetesDefaultConfigDict(TypedDict):
     airflow: NotRequired[AirflowConfigDict]
+    env: NotRequired[Sequence[KubernetesEnvConfigDict]]
+    volumes: NotRequired[Sequence[dict]]
+    volumeMounts: NotRequired[Sequence[KubernetesVolumeMountConfigDict]]
```
de-code, line 34: Should add `resources`

data_pipeline/kubernetes/kubernetes_pipeline_config.py
HazalCiplak: Same `name` but different `mountPath` is fine then?

data_pipeline/kubernetes/kubernetes_pipeline_config.py
HazalCiplak: Ok we can change it, but why do think List type wouldn't support multiple keys?

data_pipeline/kubernetes/kubernetes_pipeline_config.py
de-code: > same name but different mountPath is ok then?

Yes, I believe so. Which on its own may not be useful. But using the `subPath` would allow you to mount a select sub path from the resource specified by `name` to the `mountPath`.

data_pipeline/kubernetes/kubernetes_pipeline_config.py
de-code: I am not saying that `List` type isn't supporting it. This is just from the best practice to use the type describing what we need. And here we just need a `Sequence` (a `List` being a more specific mutable type of `Sequence`).

By "if we still want to continue supporting multiple keys" I meant:
If we just use `mountPath` as the key for volume mounts, then we don't seem to have a need for supporting multiple keys.

PR #1612, opened 2025-01-30
Remove CiviCRM Pipeline

HazalCiplak:
part of https://github.com/elifesciences/data-hub-issues/issues/1066


tests/unit_test/generic_web_api/generic_web_api_config_test.py
```diff
@@ -232,24 +231,6 @@ def test_should_read_batch_size(self):
         })
         assert web_api_config.batch_size == 123
 
-    def test_should_read_request_builder_name_and_parameters(self):
```
de-code, line 235: I think this test wasn't `civi` specific in so far that it tested request builder specific parameters. Although it was `civi` specific in that it's was only request builder that used them.

So we should probably also remove `request_builder_parameters` from `WebApiDynamicRequestBuilder` (and the corresponding support for it in the config).

PR #1632, opened 2025-02-25
Scheduled Queries pipeline

de-code:
part of https://github.com/elifesciences/data-hub-issues/issues/409

Similar to [BigQuery's Scheduled queries](https://cloud.google.com/bigquery/docs/scheduling-queries) but makes them more visible within Data Hub.


de-code:
This should be complete now for where we don't need to maintain a state file

PR #1648, opened 2025-03-24
Web API: Added support for secrets in URL

de-code:
part of https://github.com/elifesciences/data-hub-issues/issues/1099


de-code:
Most of the work so far was to prevent the password from appearing in the log (although providing an API that asks for the password to be part of the URL is really asking for it).

Currently the full URL including password still appears in the exception message.

tests/unit_test/utils/pipeline_utils_test.py
```diff
@@ -70,17 +71,6 @@ def _iter_dict_from_bq_query_mock():
         yield mock
 
 
-@pytest.fixture(name='http_error')
-def _http_error() -> HTTPError:
-    return HTTPError(
```
de-code, line 75: That HTTPError is the wrong class and we also passed in incorrect values

de-code:
I've updated the PR to reduce the scope to being able to pass in secrets to the URL (as it was required by the cited-by API, but it is not limited to). It is also including minor refactory. Maybe it could be worth to get it merged.

PR #1697, opened 2025-06-09
Add scheduled query for ga4 downloads by page and date

HazalCiplak:
related to https://github.com/elifesciences/data-hub-issues/issues/1148


sample_data_config/scheduled-queries/metrics_page_views_and_downloads_by_date.config.yaml
```diff
@@ -39,26 +40,28 @@ scheduledQueries:
...
               AND event._TABLE_SUFFIX >= FORMAT_DATE('%Y%m%d', DATE_SUB(DATE('{start_date}'), INTERVAL 3 DAY))
               AND event._TABLE_SUFFIX < FORMAT_DATE('%Y%m%d', DATE_ADD(DATE('{start_date}'), INTERVAL 1 DAY))
           )
           SELECT
             page_location,
+            event_name,
             event_date,
-            COUNT(DISTINCT ga_session_id) AS unique_page_views,
-            COUNT(*) AS page_views
+            COUNT(DISTINCT ga_session_id) AS unique_event_count,
```
de-code: Would maybe `session_count` be a more descriptive name?

sample_data_config/scheduled-queries/ga4_metrics_event_counts_by_date.config.yaml
```diff
@@ -39,26 +40,28 @@ scheduledQueries:
...
               AND event._TABLE_SUFFIX >= FORMAT_DATE('%Y%m%d', DATE_SUB(DATE('{start_date}'), INTERVAL 3 DAY))
               AND event._TABLE_SUFFIX < FORMAT_DATE('%Y%m%d', DATE_ADD(DATE('{start_date}'), INTERVAL 1 DAY))
           )
           SELECT
             page_location,
+            event_name,
             event_date,
-            COUNT(DISTINCT ga_session_id) AS unique_page_views,
-            COUNT(*) AS page_views
+            COUNT(DISTINCT ga_session_id) AS unique_event_count,
```
de-code, line 51: I had another thought on the naming of this field...

For most events other than `page_views`, it could be normal to have multiple events with the same `event_name` but different parameters. e.g. a click on two different buttons.
`file_download` is effectively a `click` event where the link target is recognised as a file download.
That is why `unique_event_count` seem to make more sense when taking all those parameters into account.
Whereas `session_count`, or `unique_session_count` if you prefer, would just say what we are counting, which seems more accurate considering the current implementation.

With most aggregates apart from raw counts we would still be aware of the breakdown dimensions.
(To counter that we could for example collect an array of session ids then we could calculate the distinct count with another breakdown - but not sure that is necessary now)

PR #1760, opened 2025-08-20
CSV pipeline: Refactor state loading

HazalCiplak:
part of https://github.com/elifesciences/data-hub-issues/issues/1192


de-code:
We now have `CsvState` which is more less handling managing the state (except for actually putting it into S3 or doing the final serialization to JSON). It doesn't include the ETag yet or a changed structure.

Things we might need to do next (just some thoughts):

- Use `CsvState` in the CSV pipeline
- To make it easier, we could consider removing the DAG (otherwise we might need to keep old functions around)
- Once we use it, we can make `CsvState` understand the current state JSON as well as the new structure
- Then we can change how it saves the state
- Then we could add ETag etc

de-code:
Found an issue not introduced by this PR that at least locally is processing the last test files even if it has processed it in the last run. Probably due to the lacking timezone. In staging I couldn't observe that.

PR #1796, opened 2025-10-01
Google Spreadsheet: Save last modified timestamp to state

de-code:
part of https://github.com/elifesciences/data-hub-issues/issues/1082

### DIFF
--- a/sample_data_config/google-spreadsheet/spreadsheet-data-pipeline.config.yaml
+++ b/sample_data_config/google-spreadsheet/spreadsheet-data-pipeline.config.yaml
@@ -5,6 +5,9 @@ spreadsheets:
   # https://docs.google.com/spreadsheets/d/1UbM9eExvJligr1bzq22xdK3cL8KIW6oqyJnu8R3uSrU/edit#gid=0
   - dataPipelineId: test_google_spreadsheet_pipeline_1
     spreadsheetId: '1UbM9eExvJligr1bzq22xdK3cL8KIW6oqyJnu8R3uSrU'
+    stateFile:
+      bucketName: '{ENV}-elife-data-pipeline'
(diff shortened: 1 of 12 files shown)


de-code:
We added optional state config, we are loading the config and are logging it.

Next we might want to then write to it.

sample_data_config/google-spreadsheet/spreadsheet-data-pipeline.config.yaml
```diff
@@ -5,6 +5,9 @@ spreadsheets:
   # https://docs.google.com/spreadsheets/d/1UbM9eExvJligr1bzq22xdK3cL8KIW6oqyJnu8R3uSrU/edit#gid=0
   - dataPipelineId: test_google_spreadsheet_pipeline_1
     spreadsheetId: '1UbM9eExvJligr1bzq22xdK3cL8KIW6oqyJnu8R3uSrU'
+    stateFile:
+      bucketName: '{ENV}-elife-data-pipeline'
+      objectName: 'airflow-config/google-spreadsheet/{ENV}-test_google_spreadsheet_pipeline_1-state.json'
```
de-code: We always seem to be using the `.json` extension for state file. Although the timestamp string isn't JSON (it would only be valid JSON if it had double quotes around it).

Maybe using something like `.txt` would be more accurate?

de-code:
This seems to be working as intended now.

Just the question regarding file extension.

sample_data_config/google-spreadsheet/spreadsheet-data-pipeline.config.yaml
HazalCiplak: yes I agree, it makes sense to update it to be txt file.

PR #1862, opened 2025-12-16
Convert Semantic SurveyMonkey pipeline to Kubernetes Pipeline

de-code:
part of https://github.com/elifesciences/data-hub-issues/issues/1275


de-code:
BTW it looks like we only need the access token from the secrets (or maybe we could get an access token via the other secrets?)

PR #1891, opened 2026-01-30
Use uv to install dependencies

de-code:
part of https://github.com/elifesciences/data-hub-issues/issues/1324

This is using [uv](https://docs.astral.sh/uv/) as a replacement for pip (using requirements txt).

It doesn't use it for the dependency management yet.


de-code:
@HazalCiplak Please try to install and use `uv` locally. Once installed, it should be faster when installing dependencies, especially when re-installing or sharing dependencies across repos.
`uv` also replaces `pyenv`.

